PR #6, opened 2019-03-14
Add mnemonic support for private keys/mdk

justicz:
This PR adds some typed helpers for the mnemonics library, so that encoding private keys and master derivation keys is an unambiguous process.

### DIFF
--- a/mnemonic/helpers.go
+++ b/mnemonic/helpers.go
@@ -1,0 +1,44 @@
+package mnemonic
+
+import (
+	"golang.org/x/crypto/ed25519"
+
+	"github.com/algorand/go-algorand-sdk/types"
+)
+
+// FromPrivateKey is a helper that converts an ed25519 private key to a
+// human-readable mnemonic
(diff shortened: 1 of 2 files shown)


mnemonic/helpers.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// ToMasterDerivationKey is a helper that converts a mnemonic directly to a
+// master derivation key
+func ToMasterDerivationKey(mnemonic string) (mdk types.MasterDerivationKey, err error) {
+	mdkBytes, err := ToKey(mnemonic)
+	if err != nil {
+		return
+	}
+	if len(mdkBytes) != len(mdk) {
```
rotemh: I believe that `MasterDerivationKey` is fixed 32 byte size

justicz: It is -- and `ToKey` should only ever return 32-byte slices. But I wanted to do this check just in case `ToKey` is modified to allow different key lengths.

PR #7, opened 2019-03-30
Updating README.md Dynamic window size, minor refactoring

jsign:
First point might be the most important:

- Changed the amount from `100` to `1000`. Why?,  `MinBalance` rule. Moreover, the example will run successfully because trx sending is async, and since `toAddr` is brand new thus having zero balance, the transaction will be kicked out of the pending pool due to  `balance X below min 1000`. The nasty thing is that the person running the example will definitely not understand why the balances don't change even if the example ran successfully.
- I chose to byte-encode a kind `note`.
- Moved the wallet name and password out of the logic. Mostly because the password was hardcoded in two different places. (Not really meaningful, but my first test of this code was with an existing wallet were changing the password in one place, not both).
- Made the block window dynamic because the prior values were guaranteed to fail. I took advantage of `txParams` to decide a reasonable window. Using +1 instead of +2 would be a tighter window but may fail if the next block is accepted too fast.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,6 +334,9 @@ const kmdToken = "42b7482737a77d9e5dffb8493ac8899db5f95cbc744d4fcffc0f1c47a6db0c
 const algodAddress = "http://localhost:8080"
 const algodToken = "6218386c0d964e371f34bbff4adf543dab14a7d9720c11c6f11970774d4575de"
 
+const walletName = "testwallet"
+const walletPassword = "testpassword"
+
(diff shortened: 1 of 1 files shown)


jsign:
Now I realize that point number one is wrong. 28 days ago a new field was added so the current example is ok!. I'll fix that in the PR soon.

jsign:
Done, also fixed the original PR description and rebased it to `master`.

PR #16, opened 2019-05-16
Add multisignature support

Vervious:
Adds support for multisignature transaction manipulation, for both payment and keyreg types, with the following API:
- `SignMultisigTransaction(sk, multisigAccount, transaction) (txid, signedBytes, error)`
- `MergeMultisigTransactions(signedBytes1, signedBytes2,...) (txid, signedBytes, error)`
- `AppendMultisigTransactions(sk, multisigAccount, signedBytesIn) (txid, signedBytes, error)`
- `MultisigAccount{Version,Threshold,Pks}` convenience type for manipulating multisig identity

The return types are `[]byte` msg-pack-encoded partially-signed multisig transactions. They can be written to files, and manipulated with `goal`, the `js-sdk`, and `algokey`.

This commit also fixes:
- serialization/deserialization of `KeyReg` transactions


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 0.1.0
```
tsachiherman: do we want to align this version with the official release version of the go-algorand ?

README.md
```diff
@@ -535,3 +535,83 @@ func main() {
...
+if err != nil {
+	panic("could not sign multisig transaction")
+}
+fmt.Printf("Made partially-signed multisig transaction with TxID %s: %x\n", txid, txBytes)
+
+```
+
+Now, we can write the returned bytes to disk:
+```golang
+file, _ := os.Create("./arbitrary_file.tx")
```
tsachiherman: nothing incorrect here, but I'm a bigger fan of `ioutil.WriteFile` which would do these three lines in one command.

README.md
```diff
@@ -535,3 +535,83 @@ func main() {
...
+_, _ := file.Write(txBytes)
+file.Close()
+```
+
+And read them back in:
+```golang
+file, _ := os.Open("./arbitrary_file.tx")
```
tsachiherman: same here -  `ioutil.ReadFile`

crypto/account.go
```diff
@@ -43,3 +44,72 @@ func GenerateAccount() (kp Account) {
...
+}
+
+// MultisigAccountFromSig is a convenience method that creates an account
+// from a sig in a signed tx. Useful for getting addresses from signed msig txs, etc.
+func MultisigAccountFromSig(sig types.MultisigSig) (ma MultisigAccount, err error) {
+	ma.Version = sig.Version
+	ma.Threshold = sig.Threshold
+	ma.Pks = make([]ed25519.PublicKey, len(sig.Subsigs))
+	for i := 0; i < len(sig.Subsigs); i++ {
+		c := make([]byte, len(sig.Subsigs[i].Key))
```
tsachiherman: is there any particular reason we're creating a copy of these keys, rather than using the input buffer ?

crypto/account.go
```diff
@@ -43,3 +44,72 @@ func GenerateAccount() (kp Account) {
...
+// Address takes this multisig preimage data, and generates the corresponding identifying
+// address, committing to the exact group, version, and public keys that it requires to sign.
+// Hash("MultisigAddr" || version uint8 || threshold uint8 || PK1 || PK2 || ...)
+func (ma MultisigAccount) Address() (addr types.Address, err error) {
+	// See go-algorand/crypto/multisig.go
+	err = ma.Validate()
+	if err != nil {
+		return
+	}
+	const msigAddrPrefix = "MultisigAddr"
```
tsachiherman: nit: I think it would be cleaner to have this defined at the top of the file, as
```const msigAddrPrefix = []byte("MultisigAddr")```
so that it could be used here directly, without any casting.

File: crypto/errors.go [hunk 6]
```diff
@@ -1,7 +1,15 @@
 package crypto
 
 import (
+	"errors"
 	"fmt"
 )
 
 var errInvalidSignatureReturned = fmt.Errorf("ed25519 library returned an invalid signature")
```
tsachiherman: If all the rest of the error are defined using `errors.New`, then this one should be defined using `errors.New` as well.

crypto/account.go
Vervious: It's a little safer, protects against the caller messing around with the keys.

CHANGELOG.md
rotemh: We should, this should already be 1.1.0 
and I think we should probably follow semantic versioning carefully across all products.

Vervious: I'm going to hold off of this until we have this conversation across all the repos. Seems like a larger decision to make.

PR #18, opened 2019-05-27
Add support for pending transactions /transactions/pending/<txid>

fabrice102:
This support is convenient to check whether a sent transaction has been committed to the ledger or not.


rotemh:
Thanks for the PR! Can you please fix the merge conflict? (by practically removing the update to the models.go file, the field is already there).

PR #20, opened 2019-05-27
Add GenesisHash to transaction header

fabrice102:


rotemh:
Hi, 
Many thanks for the PR! 
The change was actually wider than just adding the header. I fixed it in another PR. Closing this one.

PR #22, opened 2019-05-30
Fix Wrapper transactionsByAddrParams

songchenwen:
Fix Wrapper transactionsByAddrParams, make some fields omitempty.

### DIFF
--- a/client/algod/wrappers.go
+++ b/client/algod/wrappers.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"time"
 
 	"github.com/algorand/go-algorand-sdk/client/algod/models"
 )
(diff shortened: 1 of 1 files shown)


client/algod/wrappers.go
```diff
@@ -88,8 +87,7 @@ func (client Client) TransactionsByAddrForDate(addr string, first, last string)
 // TransactionsByAddr returns all transactions for a PK [addr] since the given date, with the given limit.
 // Dates are of the form "2006-01-02".
 func (client Client) TransactionsByAddrSinceDate(addr string, sinceDate string, limit uint64) (response models.TransactionList, err error) {
-	now := time.Now().Format("2006-01-02")
-	params := transactionsByAddrParams{FromDate: sinceDate, ToDate: now, Max: limit}
+	params := transactionsByAddrParams{ToDate: sinceDate, Max: limit}
```
rotemh: You've changed the meaning of the function. Please change the name to `TransactionsByAddrToDate` and the parameter `sinceDate string` to `toDate string` along with the proper comment -- 
```
// TransactionsByAddrToDate returns all transactions for a PK [addr] since the first its transaction and up to a specific date, with the given limit.
// Dates are of the form "2006-01-02".
```

Thanks!

songchenwen: Resolved

PR #26, opened 2019-06-19
add settlement and param support

jsign:
While I was hacking around the last production auction data, I stumble upon not being able to parse `Settlement` and `Params` message types.

In order to add this functionality to the SDK, I used the same structs from [here](https://github.com/algorand/go-algorand/blob/master/auction/messages.go), taking care to not use the `crypto` package for crypto types (`Digest` and `Signature`) since they're resolved in the same `type` package.

### DIFF
--- a/types/auction.go
+++ b/types/auction.go
@@ -1,5 +1,65 @@
 package types
 
+import (
+	"github.com/algorand/go-algorand/protocol"
+)
+
+// Deposit represents a deposit of external currency with a bank
+// like CoinList, for a specific auction number.
+type Deposit struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	// BidderKey is the hash of the bidder's public key, used to
+	// authenticate bids paid for by this deposit.
+	BidderKey Digest `codec:"key"`
+
+	// WinningsAddress is the address to which the winning algos should
+	// be sent to (Optional, if this field is empty, the address will be set
+	// to BidderKey)
+	WinningsAddress Digest `codec:"out"`
+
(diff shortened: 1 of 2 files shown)


jsign:
Sure, I can add all the other missing structs to complete the model.

Thinking long term, do you think that some time in the future may be useful to create other repository for the core models? So use that repo as a dependency to the Go SDK and go-algorand to avoid syncing them manually.

jsign:
Some more changes:

- Add other missing types & methods (as suggested by Nikolai)
- Indirectly noticed and completed two missing comments in public methods for Algo<->Microalgo conversion to avoid go-fmt warning.

zeldovich:
I wonder what's the right thing to do w.r.t. importing `github.com/algorand/go-algorand` (since you are now importing `github.com/algorand/go-algorand/protocol` in the auction code).  Not importing it used to be important before `go-algorand` was public.  Now, maybe it makes sense to replace more parts of the SDK code with an import of `go-algorand`.  In particular, all of the auction code could be removed from the Go SDK and replaced with importing `github.com/algorand/go-algorand/auction` instead.

Let us think about this for a bit..

jsign:
Yes, that sounds reasonable now that go-algorand is public.

Vervious:
There is currently a parallel effort to move auctions code out of the `go-algorand` in general, using the `go-sdk`, I'll keep us posted.

EvanJRichard:
@jsign I’d like to close this PR as stale, is that okay? If so, feel free to reopen or open another PR. 🙂

jsign:
Sure go ahead!

PR #28, opened 2019-06-28
add new helper MakeKeyRegTxns method

Vervious:


transaction/transaction.go
```diff
@@ -84,3 +155,15 @@ func estimateSize(txn types.Transaction) (uint64, error) {
 	}
 	return uint64(len(stx)), nil
 }
+
+func byteArrayFromBase64(in string) (out [32]byte, err error) {
```
rotemh: Consider removing this wrapper and doing it inline to match the rest of the code. If the point is to make sure it's 32 bytes, I'd change the name of the function. Also, I think it's worth using a constant instead of directly 32.

Vervious:
@27

transaction/transaction.go
Vervious: I would like to preserve the length check. Is this better?

PR #30, opened 2019-07-17
[WIP] Changes for porting auction code to go-algorand-sdk

winder:


winder:
@egieseke Does the algod REST API support that? Ideally we would be using swagger to generate this rest client to make it easier to synchronize code across SDKs, but I'm picking my battles here.

rotemh:
The API doesn't support paging and neither the SDK as a direct result

egieseke:
I created a Jira user story to extend the API and SDKs to support paging:  GOAL2-868

PR #33, opened 2019-08-02
Pst/custom token

fxgamundi:
Add support for multiple different headers - backwards compatible for existing code

### DIFF
--- a/client/algod/algod.go
+++ b/client/algod/algod.go
@@ -90,7 +90,7 @@ func stripTransaction(tx string) string {
 }
 
 // submitForm is a helper used for submitting (ex.) GETs and POSTs to the server
-func (client Client) submitForm(response interface{}, path string, request interface{}, requestMethod string, encodeJSON bool) error {
+func (client Client) submitForm(response interface{}, path string, request interface{}, requestMethod string, encodeJSON bool, headers []*Header) error {
 	var err error
 	queryURL := client.serverURL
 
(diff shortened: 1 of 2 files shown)


Vervious:
Since `go-algorand` doesn't support other api tokens this change seems orthogonal and complicates the client. Perhaps we should consider having a Pst fork?

Vervious:
Hi @fxgamundi, thoughts on us closing this pull request for the aforementioned reason? Or is there a good rationale for this change to be added?

fxgamundi:
Hi @Vervious ! I think it's a good thing for the native SDK to support
different token headers other than the built in X-Algo-API-Token. It would
make a much easier work to anyone that, like us, has a wrapper on the node
so that the calls don't go directly to it, but still have 100%
compatibility for the algod REST APIs.
Cultivating an environment conducive to the desired proliferation of Algo
tokens constitutes the call for integration with external libraries. The
allowance of secondary tokens then seems tantamount to the direction of
Algorand.

bjweaver:
Worth noting that the JS SDK supports both X-Api-Key and X-Algo-API-Token, so it would be ideal if the Go SDK did as well.

fxgamundi:
We added the new factory function to create an algod client with headers.
Also modified the client's functions accept additional headers for each specific request, making it on par with the Python SDK.
The best approach that we could think of for this was making the functions variadic, with the headers as the last arg of the functions. This guaranties that the functions can be called without any headers, just one, or a slice of them, making it fully backwards compatible.

PR #41, opened 2019-08-09
Add signing and verifying arbitrary bytes

mjiang102628:


crypto/crypto.go
```diff
@@ -17,6 +17,9 @@ var txidPrefix = []byte("TX")
 // bidPrefix is prepended to a bid when signing it
 var bidPrefix = []byte("aB")
 
+// bytesPrefix is prepended to bytes when signing
+var bytesPrefix = []byte("NF")
```
zeldovich: It's important to keep track of these prefixes in `protocol/hash.go` in the `go-algorand` repo (for lack of a better authoritative place), so that our signed messages are always unambiguous.

What's the intended use for this kind of signed message?

How does it differ from the `MX` prefix we already have for free-form messages (the `Message` HashID value)?

PR #45, opened 2019-08-22
add note parameter to MakeKeyRegTxn

mjiang102628:


transaction/transaction_test.go
```diff
@@ -93,7 +93,7 @@ func TestKeyRegTxn(t *testing.T) {
 
 func TestMakeKeyRegTxn(t *testing.T) {
```
EvanJRichard: Do you think the magic numbers in this test (`45, 67`, `3060`) need commenting, or no?

mjiang102628: I don't think so; 45 and 67 are arbitrary bytes for the note field and 3060 is just the expected fee after accounting for size.

PR #47, opened 2019-09-12
Make asset creation txns

EvanJRichard:
## Summary
Add `MakeUnsignedAssetCreateTx`, modeled after `libgoal`'s helper of the same name.

### Tests
Added a unit test to `transaction_test.go`. I have a branch in progress on `algorand-sdk-testing` for the cross-sdk testing for this feature, it will be merged before the next feature begins.


transaction/transaction.go
```diff
@@ -195,6 +194,119 @@ func MakeKeyRegTxnWithFlatFee(account string, fee, firstRound, lastRound uint64,
 	return tx, nil
 }
 
+// MakeAssetCreateTxn constructs an asset creation transaction using the passed parameters.
+// - account is a checksummed, human-readable address for which we register the given participation key.
```
rotemh: Participation key?

transaction/transaction.go
```diff
@@ -195,6 +194,119 @@ func MakeKeyRegTxnWithFlatFee(account string, fee, firstRound, lastRound uint64,
...
+// - fee is fee per byte as received from algod SuggestedFee API call.
+// - firstRound is the first round this txn is valid (txn semantics unrelated to key registration)
```
rotemh: same here and the other function

PR #49, opened 2019-09-13
Add transactions group support

pzbitskiy:
* New ComputeGroupID return group ID for group of transactions
* TestComputeGroupID verified group ID agains 'goal clerk group' output
* README contains usage example

Note, no special SendTransaction function added because SendRawTransaction works great and there not API to get SignedTxn object from signing functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -610,3 +610,92 @@ We can also merge raw, partially-signed multisig transactions:
 otherTxBytes := ... // generate another raw multisig transaction somehow
 txid, mergedTxBytes, err := crypto.MergeMultisigTransactions(twoOfThreeTxBytes, otherTxBytes)
 ```
+
+## Working with transactions group
+
+Example below demonstrates how to create a transactions group and send it to network.
+
+```golang
+package main
+
+import (
+	"fmt"
+
+	"github.com/algorand/go-algorand-sdk/client/algod"
+	"github.com/algorand/go-algorand-sdk/crypto"
+	"github.com/algorand/go-algorand-sdk/transaction"
+	"github.com/algorand/go-algorand-sdk/types"
+)
+
+// CHANGE ME
+const algodAddress = "http://localhost:8080"
+const algodToken = "f1dee49e36a82face92fdb21cd3d340a1b369925cd12f3ee7371378f1665b9b1"
+
+func submitGroup() {
+	account1 := crypto.GenerateAccount()
+	fmt.Printf("account address: %s\n", account1.Address)
+	account2 := crypto.GenerateAccount()
+	fmt.Printf("account address: %s\n", account2.Address)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -614,3 +614,92 @@ We can also merge raw, partially-signed multisig transactions:
...
+		address2, address3, fee, amount2, firstRound2, firstRound2+1000,
+		note, "", genesisID, genesisHash,
+	)
+	if err != nil {
+		fmt.Printf("Failed to create payment transaction: %v\n", err)
+		return
+	}
+
+	// compute group id and put it into each transaction
+	gid, err := crypto.ComputeGroupID([]types.Transaction{tx1, tx2})
```
rotemh: Please consider adding the same `AssignGroupID` to keep it consistent with the Py SDK?

crypto/crypto.go
```diff
@@ -59,9 +62,14 @@ func rawTransactionBytesToSign(tx types.Transaction) ([]byte) {
 	return bytes.Join(msgParts, nil)
 }
 
+// rawHash computes Sum512_256 hash from raw bytes
+func rawHash(input []byte) types.Digest {
```
rotemh: Since we really use only one hash function, I believe that there is no need to wrap it.

types/transaction.go
```diff
@@ -61,4 +61,21 @@ type Header struct {
...
+// TxGroup describes a group of transactions that must appear
+// together in a specific order in a block.
+type TxGroup struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	// Transactions specifies a list of transactions that must appear
+	// together, sequentially, in a block in order for the group to be
+	// valid.  Each hash in the list is a hash of a transaction with
+	// the `Group` field omitted.
+	Transactions []Digest `codec:"txlist"`
```
rotemh: Although I believe it was copied from our main repo, I think that the name `TxGroup` is a bit misleading. I expected to see `Transaction []Transaction` but instead it contains the hashes as `Digest`. Maybe we should consider changing the name to `TxGroupHahses`? and the comment accordingly? @zeldovich - Was there a reason for name that I missing?

zeldovich: I don't think there's anything deep going on here.  Renaming it seems fine.  I think consistency between the SDK and the `go-algorand` codebase is important, so that means renaming it in `go-algorand` too if you care enough about renaming.

pzbitskiy: I'll rename here and there.

README.md
pzbitskiy: Done

crypto/crypto.go
pzbitskiy: Done

PR #55, opened 2019-09-26
Destroy-asset txns 

EvanJRichard:
## Summary
This PR proposes to add support for making a destroy-asset transaction, which removes an asset from the record entirely. To destroy an asset, the asset creator must own all of the outstanding issued asset, and the destruction transaction must be issued by the asset manager. The destruction transaction is as a configuration transaction, but with `AssetParams` zeroed.

### Testing
See `transaction_test.go`

### See also
Working on equivalent PRs for python, js, java.


transaction/transaction.go
```diff
@@ -281,9 +281,9 @@ func MakeAssetCreateTxn(account string, feePerByte, firstRound, lastRound uint64
 // keys for an asset. An empty string means a zero key (which
 // cannot be changed after becoming zero); to keep a key
 // unchanged, you must specify that key.
-// - account is a checksummed, human-readable address for which we register the given participation key.
-// - fee is a flat fee
```
pzbitskiy: good catch

transaction/transaction.go
```diff
@@ -405,6 +429,22 @@ func MakeAssetConfigTxnWithFlatFee(account string, fee, firstRound, lastRound ui
 	return tx, nil
 }
 
+// MakeAssetDestroyTxn creates a tx template for destroying an asset, removing it from the record.
+// All outstanding asset amount must be held by the creator, and this transaction must be issued by the asset manager.
+// - account is a checksummed, human-readable address that will send the transaction; it also must be the asset manager
+// - fee is a fee per byte
```
pzbitskiy: flat fee

EvanJRichard:
@pzbitskiy :+1: Addressed incorrect code comment, good eyes. Re: point 2, the end to end testing - the `go-algorand-sdk` has no end to end testing or testing against `goal`, as far as I can tell. But, adding asset tests to `algorand-sdk-testing` is on the roadmap. Is that sufficient?

pzbitskiy:
> @pzbitskiy 👍 Addressed incorrect code comment, good eyes. Re: point 2, the end to end testing - the `go-algorand-sdk` has no end to end testing or testing against `goal`, as far as I can tell. But, adding asset tests to `algorand-sdk-testing` is on the roadmap. Is that sufficient?

I thought we have kind of e2e test - all these golden serializations and de-serializations

PR #60, opened 2019-10-08
Asset Transfer Transactions, and related operations: acceptance, revocation, burning, minting

EvanJRichard:
## Summary
This PR proposes to add asset transfer transactions.
Additionally, it proposes to add "I accept asset X" transactions, which are a special case of asset transfer transactions.
Additionally, it proposes to add revocation transactions, which are asset transfer transactions where the clawback address is revoking asset ownership from another account. 
Additionally, it proposes to add helpers for mint and burn asset transactions. N.B. these helpers do not query the asset's AssetParams to see if the `reserve` account passed to the helper actually map to the `reserve` account stored for the asset.

### Testing
Unit tests in `transaction_test.go` WIP

### DIFF
--- a/transaction/transaction.go
+++ b/transaction/transaction.go
@@ -285,6 +285,7 @@ func MakeAssetCreateTxn(account string, feePerByte, firstRound, lastRound uint64
 // - fee is a fee per byte
 // - firstRound is the first round this txn is valid (txn semantics unrelated to asset config)
 // - lastRound is the last round this txn is valid
+// - note is an arbitrary byte array
 // - genesis id corresponds to the id of the network
 // - genesis hash corresponds to the base64-encoded hash of the genesis of the network
 func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64, note []byte, genesisID string, genesisHash string, creator string,
@@ -310,6 +311,7 @@ func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64
 		LastValid:   types.Round(lastRound),
 		GenesisHash: types.Digest(ghBytes),
 		GenesisID:   genesisID,
+		Note:        note,
 	}
 
 	creatorAddr, err := types.DecodeAddress(creator)
@@ -370,6 +372,7 @@ func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64
 // - feePerByte is a fee per byte
 // - firstRound is the first round this txn is valid (txn semantics unrelated to asset management)
 // - lastRound is the last round this txn is valid
+// - note is an arbitrary byte array
 // - genesis id corresponds to the id of the network
 // - genesis hash corresponds to the base64-encoded hash of the genesis of the network
 // - creator is the address of the asset creator
@@ -397,6 +400,7 @@ func MakeAssetTransferTxn(account, recipient, closeAssetsTo string, amount, feeP
 		LastValid:   types.Round(lastRound),
 		GenesisHash: types.Digest(ghBytes),
 		GenesisID:   genesisID,
+		Note:        note,
 	}
 
 	creatorAddr, err := types.DecodeAddress(creator)
@@ -440,6 +444,7 @@ func MakeAssetTransferTxn(account, recipient, closeAssetsTo string, amount, feeP
 // - feePerByte is a fee per byte
 // - firstRound is the first round this txn is valid (txn semantics unrelated to asset management)
 // - lastRound is the last round this txn is valid
+// - note is an arbitrary byte array
 // - genesis id corresponds to the id of the network
 // - genesis hash corresponds to the base64-encoded hash of the genesis of the network
 // - creator is the address of the asset creator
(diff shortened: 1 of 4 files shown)


transaction/transaction.go
```diff
@@ -361,6 +361,96 @@ func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64
...
+// MakeAssetTransferTxn creates a tx for sending some asset from an asset holder to another user
+// the recipient must have marked themselves as willing to accept the asset
+// - account is a checksummed, human-readable address that will send the transaction and assets
+// - recipient is a checksummed, human-readable address what will receive the assets
+// - closeAssetsTo is a checksummed, human-readable address that behaves as a close-to address for the asset transaction; the remaining assets not sent to recipient will be sent to closeAssetsTo. Leave blank for no close-to behavior.
+// - amount is the number of assets to send
+// - feePerByte is a fee per byte
+// - firstRound is the first round this txn is valid (txn semantics unrelated to asset management)
+// - lastRound is the last round this txn is valid
+// - genesis id corresponds to the id of the network
```
algonautshant: note is missing?

EvanJRichard: @algonautshant yeah, good find!

transaction/transaction.go
```diff
@@ -361,6 +361,96 @@ func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64
...
+
+	tx.AssetAmount = amount
+
+	// Update fee
+	eSize, err := estimateSize(tx)
+	if err != nil {
+		return types.Transaction{}, err
+	}
+	tx.Fee = types.MicroAlgos(eSize * feePerByte)
+
```
algonautshant: Why is this not checked against MinTxnFee?

EvanJRichard: It should be 😀adding check against this everywhere it seems missing

transaction/transaction.go
```diff
@@ -361,6 +363,132 @@ func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64
...
+	return tx, nil
+}
+
+// MakeAssetAcceptanceTransaction creates a tx for marking an asset as willing to be accepted by an account
```
algonautshant: Is the marking the asset or marking the account? 
This statement does not seem to agree with the other comment: 
"the recipient must have marked themselves as willing to accept the asset"

transaction/transaction.go
```diff
@@ -361,6 +363,132 @@ func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64
 	return tx, nil
 }
 
+// MakeAssetTransferTxn creates a tx for sending some asset from an asset holder to another user
+// the recipient must have marked themselves as willing to accept the asset
```
algonautshant: itself?

transaction/transaction.go
```diff
@@ -490,7 +685,148 @@ func MakeAssetConfigTxnWithFlatFee(account string, fee, firstRound, lastRound ui
...
+// - amount is the number of assets to send
+// - fee is the flat fee
+// - firstRound is the first round this txn is valid (txn semantics unrelated to asset management)
+// - lastRound is the last round this txn is valid
+// - note is an arbitrary byte array
+// - genesis id corresponds to the id of the network
+// - genesis hash corresponds to the base64-encoded hash of the genesis of the network
+// - creator is the address of the asset creator
+// - index is the asset index
+func MakeAssetMintTransactionWithFlatFee(reserve, recipient string, amount, fee, firstRound, lastRound uint64, note []byte,
```
mjiang102628: Maybe have this wrap around MakeAssetTransferTxnWithFlatFee instead? Then you don't need to check for flat fee since MakeAssetTransferTxnWithFlatFee takes care of it.

transaction/transaction.go
```diff
@@ -285,6 +291,7 @@ func MakeAssetCreateTxn(account string, feePerByte, firstRound, lastRound uint64
 // - fee is a fee per byte
 // - firstRound is the first round this txn is valid (txn semantics unrelated to asset config)
 // - lastRound is the last round this txn is valid
+// - note is an arbitrary byte array
 // - genesis id corresponds to the id of the network
 // - genesis hash corresponds to the base64-encoded hash of the genesis of the network
 func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64, note []byte, genesisID string, genesisHash string, creator string,
```
rotemh: Is `MakeAssetConfigTxn` intended to change the configuration on an existing asset? If so, please consider enhancing the comment.

transaction/transaction.go
```diff
@@ -358,9 +366,187 @@ func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64
...
+// - creator is the address of the asset creator
+// - index is the asset index
+func MakeAssetAcceptanceTxn(account string, feePerByte, firstRound, lastRound uint64, note []byte,
+	genesisID, genesisHash, creator string, index uint64) (types.Transaction, error) {
+	tx, err := MakeAssetTransferTxn(account, account, "", 0,
+		feePerByte, firstRound, lastRound, note, genesisID, genesisHash, creator, index)
+	return tx, err
+}
+
+// MakeAssetRevocationTransaction creates a tx for revoking an asset from an account and sending it to another
```
rotemh: Do we actually need this function? Practically, just the revocation manager can revoke assets so it seems a bit redundant to the regular transfer asset function

transaction/transaction.go
```diff
@@ -358,9 +366,187 @@ func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64
...
+
+	tx.Fee = types.MicroAlgos(eSize * feePerByte)
+	if tx.Fee < MinTxnFee {
+		tx.Fee = MinTxnFee
+	}
+
+	return tx, err
+}
+
+// MakeAssetMintTransaction creates a tx for minting more of an asset (transferring from the reserve account)
```
rotemh: Same for this one. I believe it would be much easier to simply explain this in the documentation but it worth a discussion.

transaction/transaction.go
```diff
@@ -358,9 +366,187 @@ func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64
...
+// - amount is the number of assets to "burn"
+// - feePerByte is a fee per byte
+// - firstRound is the first round this txn is valid (txn semantics unrelated to asset management)
+// - lastRound is the last round this txn is valid
+// - note is an arbitrary byte array
+// - genesis id corresponds to the id of the network
+// - genesis hash corresponds to the base64-encoded hash of the genesis of the network
+// - creator is the address of the asset creator
+// - index is the asset index
+func MakeAssetBurnTxn(account, reserve string, amount, feePerByte, firstRound, lastRound uint64, note []byte,
```
rotemh: Same here. The rational behind removing these functions is that they no real added value. If the functions were to compute or pull the relevant information automatically (such as the reserve address and so on), I'd agree, but right now, I think it may be more confusing to the user. Again, worth discussion.

transaction/transaction.go
EvanJRichard: @rotemh I think it's worth it to keep the revocation helper, since it adds handling `tx.AssetSend`/`asnd`, which normal transfer doesn't need to worry about. I agree that `Mint` or `Burn` don't add anything, I'll remove them.

EvanJRichard: Follow-up: nevermind, leaving them for now per our in-person conversation

Issue #67, opened 2019-10-22
golint violations in transaction.go


EvanJRichard:
```
pzbitskiy@mbr: .../go-algorand-sdk (pavel/logicsig)$ make 
golint `go list ./... | grep -v /vendor/`
algorand/go-algorand-sdk/transaction/transaction.go:462:1: comment on exported function MakeAssetAcceptanceTxn should be of the form "MakeAssetAcceptanceTxn ..."
algorand/go-algorand-sdk/transaction/transaction.go:479:1: comment on exported function MakeAssetRevocationTxn should be of the form "MakeAssetRevocationTxn ..."
algorand/go-algorand-sdk/transaction/transaction.go:683:1: comment on exported function MakeAssetAcceptanceTxnWithFlatFee should be of the form "MakeAssetAcceptanceTxnWithFlatFee ..."
algorand/go-algorand-sdk/transaction/transaction.go:699:1: comment on exported function MakeAssetRevocationTxnWithFlatFee should be of the form "MakeAssetRevocationTxnWithFlatFee ..."
```
found by pzbitskiy 

related thought - maybe we could have the SDKs present a reminder to `golint` to PR-ing users, like `go-algorand` PRs

PR #70, opened 2019-10-24
Update for new asset behavior (incl. README updates)

EvanJRichard:
## Summary
recent changes in `go-algorand` did things like: add support for `URL`s and `MetadataHash`es to assets. Also, the `creator` is no longer needed to index into assets. This PR proposes to update the `go` lang SDK to support these changes.
The `README.md` is also updated to discuss Assets.

### Testing
Unit tests updated.

###  See also
https://github.com/algorand/go-algorand-sdk/issues/63
https://github.com/algorand/go-algorand-sdk/issues/64
https://github.com/algorand/go-algorand-sdk/issues/67

### DIFF
--- a/client/algod/models/models.go
+++ b/client/algod/models/models.go
@@ -145,6 +145,19 @@ type AssetParams struct {
 	// required: false
 	AssetName string `json:"assetname"`
 
+	// URL specifies a URL where more information about the asset can be
+	// retrieved
+	//
+	// required: false
+	URL string `json:"url"`
+
+	// MetadataHash specifies a commitment to some unspecified asset
+	// metadata. The format of this metadata is up to the application.
+	//
+	// required: false
+	// swagger:strfmt byte
+	MetadataHash []byte `json:"metadatahash"`
+
 	// ManagerAddr specifies the address used to manage the keys of this
 	// asset and to destroy it.
 	//
--- a/transaction/transaction.go
+++ b/transaction/transaction.go
@@ -381,23 +381,11 @@ func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64
 	return tx, nil
 }
 
-// MakeAssetTransferTxn creates a tx for sending some asset from an asset holder to another user
-// the recipient address must have previously issued an asset acceptance transaction for this asset
-// - account is a checksummed, human-readable address that will send the transaction and assets
-// - recipient is a checksummed, human-readable address what will receive the assets
-// - closeAssetsTo is a checksummed, human-readable address that behaves as a close-to address for the asset transaction; the remaining assets not sent to recipient will be sent to closeAssetsTo. Leave blank for no close-to behavior.
-// - amount is the number of assets to send
-// - feePerByte is a fee per byte
-// - firstRound is the first round this txn is valid (txn semantics unrelated to asset management)
-// - lastRound is the last round this txn is valid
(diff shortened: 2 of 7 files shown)


EvanJRichard:
_@rotemh wrote:_
> Does this PR contain the new goldens?

Yes. As part of consolidating all the goldens I'm going to double check/cross-reference/etc. but these should be right. :+1:

transaction/transaction.go
```diff
@@ -326,15 +336,7 @@ func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64
...
-	}
-
-	tx.ConfigAsset = types.AssetID{
-		Creator: creatorAddr,
-		Index:   index,
-	}
+	tx.ConfigAsset = types.AssetIndex(index)
 
 	if newManager != "" {
 		tx.Type = types.AssetConfigTx
```
justicz: I believe this line is not necessary

transaction/transaction.go
```diff
@@ -297,10 +308,9 @@ func MakeAssetCreateTxn(account string, feePerByte, firstRound, lastRound uint64
 // - note is an arbitrary byte array
 // - genesis id corresponds to the id of the network
 // - genesis hash corresponds to the base64-encoded hash of the genesis of the network
-// - creator the address of the asset creator
 // - index is the asset index id
 // - for newManager, newReserve, newFreeze, newClawback see asset.go
-func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64, note []byte, genesisID string, genesisHash string, creator string,
+func MakeAssetConfigTxn(account string, feePerByte, firstRound, lastRound uint64, note []byte, genesisID, genesisHash string,
```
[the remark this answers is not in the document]
EvanJRichard: Looks like they can't be changed, so I'll leave them out here.

transaction/transaction.go
EvanJRichard: It is necessary - without it, compilation fails with `cannot use index (type uint64) as type types.AssetIndex in assignment`

client/algod/models/models.go
```diff
@@ -145,6 +145,19 @@ type AssetParams struct {
...
+	//
+	// required: false
+	URL string `json:"url"`
+
+	// MetadataHash specifies a commitment to some unspecified asset
+	// metadata. The format of this metadata is up to the application.
+	//
+	// required: false
+	// swagger:strfmt byte
+	MetadataHash []byte `json:"metadatahash"`
```
rotemh: Note that these have changed a bit (nothing too extreme). refer to https://github.com/algorand/go-algorand/blob/master/data/basics/userBalance.go#L206

transaction/transaction_test.go
```diff
@@ -132,8 +132,10 @@ func TestMakeAssetCreateTxn(t *testing.T) {
 	const clawback = addr
 	const unitName = "tst"
 	const assetName = "testcoin"
+	const testURL = "website"
+	const metadataHash = "fACPO4nRgO55j1ndAK3W6Sgc4APkcyFh"
```
rotemh: is that a real 32byte hash?
While the code should not reject a shorter than 32 bytes string (If someone decides to make another use of the field, I'm ok with it), it should reject longer ones. 
Also, for the sake of completeness, if this is not a 32 byte string, I'd use a real one.

PR #73, opened 2019-10-29
Make certain asset fields variable length

justicz:
Asset name, asset unit name, and asset URL should be variable length


EvanJRichard:
Some overlap with #70 , of note I don't think `MakeAssetConfigTxn` needs updating. Will incorporate the `MaxLen` checking into #70

justicz:
Hey @winder -- this is for the `MobileCompatible` branch

Issue #76, opened 2019-11-06
makeTxn methods for assets use string for genesisHash instead of bytes 


mjiang102628:
related to #39

mjiang102628:
Investigated: this is still an issue in the old transaction.go, with makePaymentTxn taking []byte and others taking string, but is fixed in future/transaction.go by using suggested parameters. Changes to the old functions would be backwards incompatible anyway, so closing this issue.

Issue #39, opened 2019-08-07
inconsistent genesisHash format between MakePaymentTxn and MakeKeyRegTxn


mjiang102628:
MakePaymentTxn takes the genesis hash as bytes, and MakeKeyRegTxn takes a base64 string.

EvanJRichard:
Closing - we will get this for free as part of suggested params object task

PR #77, opened 2019-11-06
Fix inconsistent genesis hash formats

mjiang102628:
This is backwards incompatible, so it will fail testing repo tests.
#76 and #39


EvanJRichard:
Closing due to eventual `Params` redesign

PR #83, opened 2019-11-13
LogicSig Template: Split, HTLC

EvanJRichard:
## Summary
This PR describes template classes for building and using LogicSig contracts. At time of writing it includes two classes, `Split` and `HTLC`.
These templates are in a new package, `templates`. They were originally in `logic`, but this caused circular imports. (The `templates` require `crypto`, which requires `logic`.)

### Tests
Golden tests in `templates_test.go`

### See also
#81


templates/template.go
```diff
@@ -0,0 +1,47 @@
...
+	"fmt"
+	"github.com/algorand/go-algorand-sdk/types"
+)
+
+func replace(buf, newBytes []byte, offset, placeholderLength uint64) []byte {
+	output := append(buf[:offset], newBytes...)
+	return append(output, buf[(offset+placeholderLength):]...)
+}
+
+func inject(original []byte, offsets []uint64, values []interface{}) (result []byte, err error) {
```
EvanJRichard: am especially concerned with whether this function is correct

templates/template.go
```diff
@@ -0,0 +1,47 @@
...
+		return
+	}
+
+	for i, value := range values {
+		decodedLength := 0
+
+		if valueAsUint, ok := value.(uint64); ok {
+			buffer := make([]byte, 8)
+			decodedLength = binary.PutUvarint(buffer, valueAsUint)
+			result = replace(result, buffer, offsets[i], uint64(decodedLength)) // TODO ejr should the last arg here just always be 1?
```
EvanJRichard: wonder if last arg of `replace` is correct here

templates/template.go
```diff
@@ -0,0 +1,47 @@
...
+	if len(offsets) != len(values) {
+		err = fmt.Errorf("length of offsets %v does not match length of replacement values %v", len(offsets), len(values))
+		return
+	}
+
+	for i, value := range values {
+		decodedLength := 0
+
+		if valueAsUint, ok := value.(uint64); ok {
+			buffer := make([]byte, 8)
```
EvanJRichard: largest # of bytes a uint64 could be is 8, so `8` is appropriate here, yeah?

rotemh: no, it can actually be larger. The encoding is different the "normal" one.

templates/template.go
rotemh: that should not be the decoded length but the place holder length that you've left in the original `referenceAsBytes` array. I think that you have them in different lengths. Might make it more difficult for you. One byte should be sufficient.

templates/template.go
EvanJRichard: What is its max size?

rotemh: `binary.MaxVarintLen64`

templates/template.go
```diff
@@ -0,0 +1,73 @@
...
+			copy(addressBytes, address[:])
+			result = replace(result, addressBytes, offsets[i], addressLen)
+		} else if b64string, ok := value.(string); ok {
+			// always assume a string is a len-32 b64string replaceing a len-32 b64 string
+			decodeBytes, decodeErr := base64.StdEncoding.DecodeString(b64string)
+			if decodeErr != nil {
+				err = decodeErr
+				return
+			}
+			result = replace(result, decodeBytes, offsets[i], uint64(32))
```
algonautshant: Shouldn't the last argument be uint64(2)?

templates/template.go
```diff
@@ -0,0 +1,73 @@
...
+			fillingBuffer := make([]byte, decodedLength)
+			decodedLength = binary.PutUvarint(fillingBuffer, valueAsUint)
+			result = replace(result, fillingBuffer, offsets[i], uint64(1))
+		} else if address, ok := value.(types.Address); ok {
+			addressLen := uint64(32)
+			addressBytes := make([]byte, addressLen)
+			copy(addressBytes, address[:])
+			result = replace(result, addressBytes, offsets[i], addressLen)
+		} else if b64string, ok := value.(string); ok {
+			// always assume a string is a len-32 b64string replacing a len-32 b64 string
```
algonautshant: Is this doing the same thing as the following? 
		elif val_type == Types.BASE64:
			val = bytes(base64.b64decode(val))
			buf = []
			dec_len = **put_uvarint**(buf, len(val)) + len(val)
			res = replace(res, bytes(buf) + val, offsets[i], 2)

PR #86, opened 2019-11-18
Change leases to be backwards compatible. Revert old lease impl.

EvanJRichard:
## Summary
Previous lease implementation broke backwards compatibility. Now, `AddLease` is used to add a lease to an existing transaction.

### Testing
Golden still passes.

### See also
https://github.com/algorand/go-algorand-sdk/issues/79

### DIFF
--- a/transaction/transaction.go
+++ b/transaction/transaction.go
@@ -64,7 +64,7 @@ func MakePaymentTxn(from, to string, fee, amount, firstRound, lastRound uint64, 
 	}
 
 	// Update fee
-	eSize, err := estimateSize(tx)
+	eSize, err := tx.EstimateSize()
 	if err != nil {
 		return types.Transaction{}, err
 	}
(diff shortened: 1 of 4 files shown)


transaction/transaction.go
```diff
@@ -778,6 +757,42 @@ func AssignGroupID(txns []types.Transaction, account string) (result []types.Tra
 	return result, nil
 }
 
+// AddLease adds the passed lease (see types/transaction.go) to the header of the passed transaction
+// - txn: the types.Transaction to modify
+// - lease: the [32]byte lease to add to the header
+// - feePerByte: the new feePerByte
+func AddLease(tx types.Transaction, lease [32]byte, feePerByte uint64) (types.Transaction, error) {
```
rotemh: Consider making these two a method of `Transaction`

EvanJRichard: @rotemh Just tried implementing that, and I liked how it looked, but it causes an import cycle that I don't see a way around. Doing that either imports `crypto` or `transaction` into the `types` package, but `crypto` and `transaction` import `types.`

PR #89, opened 2019-11-20
Refuse to compute group id for too-large groups

EvanJRichard:
https://github.com/algorand/go-algorand-sdk/issues/85


crypto/crypto.go
```diff
@@ -363,6 +363,12 @@ func VerifyMultisig(addr types.Address, message []byte, msig types.MultisigSig)
 
 // ComputeGroupID returns group ID for a group of transactions
 func ComputeGroupID(txgroup []types.Transaction) (gid types.Digest, err error) {
+	// MaxTxGroupSize is max number of transactions in a single group
+	const MaxTxGroupSize = 16
```
pzbitskiy: move the constant to types.basic?

PR #101, opened 2019-12-06
Limit order TEAL template

EvanJRichard:
## Summary
This template implements `limit-order-a.teal.impl` from `go-algorand`.

### Testing

### See also
https://github.com/algorand/go-algorand-sdk/issues/98

### DIFF
--- a/templates/limitOrder.go
+++ b/templates/limitOrder.go
@@ -18,9 +18,14 @@ type LimitOrder struct {
 // assetAmount: amount of assets to be sent
 // contract: byteform of the contract from the payer
 // secretKey: secret key for signing transactions
+// fee: fee per byte used for the transactions
+// algoAmount: number of algos to transfer
+// firstRound: first round on which these txns will be valid
+// lastRound: last round on which these txns will be valid
+// genesisHash: genesisHash indicating the network for the txns
 // the first payment sends money (Algos) from contract to the recipient (we'll call him Buyer), closing the rest of the account to Owner
 // the second payment sends money (the asset) from Buyer to the Owner
-// these transactions will be rejected if they do not meet or beat the ratio set at contract creation
+// these transactions will be rejected if they do not meet the restrictions set by the contract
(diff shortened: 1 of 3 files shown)


templates/limitOrder.go
```diff
@@ -0,0 +1,126 @@
...
+
+	var signedGroup []byte
+	signedGroup = append(signedGroup, assetsForAlgosSigned...)
+	signedGroup = append(signedGroup, algosForAssetsSigned...)
+
+	return signedGroup, nil
+}
+
+// MakeLimitOrder allows a user to exchange some number of N assets for D algos.
+// This is a contract account.
```
rotemh: This is a bit of a weird comment here that makes much more sense in the TEAL code.

templates/limitOrder.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	var signedGroup []byte
+	signedGroup = append(signedGroup, assetsForAlgosSigned...)
+	signedGroup = append(signedGroup, algosForAssetsSigned...)
+
+	return signedGroup, nil
+}
+
+// MakeLimitOrder allows a user to exchange some number of N assets for D algos.
```
rotemh: This line and line 75 are contradicting

templates/limitOrder.go
```diff
@@ -0,0 +1,125 @@
...
+// secretKey: secret key for signing transactions
+// fee: fee per byte used for the transactions
+// algoAmount: number of algos to transfer
+// firstRound: first round on which these txns will be valid
+// lastRound: last round on which these txns will be valid
+// genesisHash: genesisHash indicating the network for the txns
+// the first payment sends money (Algos) from contract to the recipient (we'll call him Buyer), closing the rest of the account to Owner
+// the second payment sends money (the asset) from Buyer to the Owner
+// these transactions will be rejected if they do not meet the restrictions set by the contract
+func (lo LimitOrder) GetSwapAssetsTransaction(assetAmount uint64, contract, secretKey []byte, fee, algoAmount, firstRound, lastRound uint64, genesisHash []byte) ([]byte, error) {
```
rotemh: This one does the correct thing!

winder: we should probably rename `algoAmount` to `microAlgoAmount`

Issue #99, opened 2019-12-04
Implement Periodic Payment Contract


EvanJRichard:
#96 #98

PR #103, opened 2019-12-13
Periodic payment TEAL template

EvanJRichard:
## Summary
This template implements `periodic-payment.teal.impl` from `go-algorand`.

### Testing

### See also
https://github.com/algorand/go-algorand-sdk/issues/99


templates/periodicPayment.go
```diff
@@ -0,0 +1,99 @@
...
+	receiver    types.Address
+	leaseBase64 string
+}
+
+// GetWithdrawalTransaction returns a signed transaction extracting funds from the contract
+// fee: the fee to pay in microAlgos
+// firstValid: the first round on which the txn will be valid
+// lastValid: the final round on which the txn will be valid
+// genesisHash: the hash representing the network for the txn
+func (c PeriodicPayment) GetWithdrawalTransaction(fee, firstValid, lastValid uint64, genesisHash []byte) ([]byte, error) {
```
rotemh:
```
&&
txn FirstValid
int TMPL_PERIOD
%
int 0
==
&&
txn LastValid
int TMPL_DUR
txn FirstValid
+
==
```


This is from the contract. `FirstValid` must be a multiple of `period` and `LastValid` must be equal to `duration + firstValid` so worth checking that and computing `LastValid` ourselves

rotemh:
I believe that the summary contains a mistake about which contract is implemented here :)

templates/periodicPayment.go
```diff
@@ -0,0 +1,108 @@
...
+//  - maxFee: maximum fee used by the withdrawal transaction
+func MakePeriodicPayment(receiver string, amount, withdrawWindow, period, expiryRound, maxFee uint64) (PeriodicPayment, error) {
+	leaseBytes := make([]byte, 32)
+	crypto.RandomBytes(leaseBytes)
+	leaseString := base64.StdEncoding.EncodeToString(leaseBytes)
+	return MakePeriodicPaymentWithLease(receiver, leaseString, amount, withdrawWindow, period, expiryRound, maxFee)
+}
+
+// MakePeriodicPaymentWithLease is as MakePeriodicPayment, but the caller can specify the lease (using b64 string)
+func MakePeriodicPaymentWithLease(receiver, lease string, amount, withdrawWindow, period, expiryRound, maxFee uint64) (PeriodicPayment, error) {
```
rotemh: The user can't choose his/er lease. It must match the original so maybe this function needs to be private?

templates/periodicPayment.go
```diff
@@ -0,0 +1,108 @@
...
+	leaseBase64 string
+	period      uint64
+	duration    uint64
+}
+
+// GetWithdrawalTransaction returns a signed transaction extracting funds from the contract
+// fee: the fee to pay in microAlgos
+// firstValid: the first round on which the txn will be valid
+// genesisHash: the hash representing the network for the txn
+func (c PeriodicPayment) GetWithdrawalTransaction(fee, firstValid uint64, genesisHash []byte) ([]byte, error) {
```
rotemh: Same like the others. Does the function expect the user to re-instantiate the contract?

PR #106, opened 2019-12-19
Dynamic fee TEAL template

EvanJRichard:
## Summary
This template implements `dynamic-fee.teal.impl` from `go-algorand`.

### Testing

### See also
https://github.com/algorand/go-algorand-sdk/issues/96

### DIFF
--- a/templates/dynamicFee.go
+++ b/templates/dynamicFee.go
@@ -2,7 +2,9 @@ package templates
 
 import (
 	"encoding/base64"
+	"fmt"
 	"github.com/algorand/go-algorand-sdk/crypto"
+	"github.com/algorand/go-algorand-sdk/logic"
 	"github.com/algorand/go-algorand-sdk/transaction"
 	"github.com/algorand/go-algorand-sdk/types"
 	"golang.org/x/crypto/ed25519"
@@ -11,12 +13,6 @@ import (
 // DynamicFee template representation
 type DynamicFee struct {
 	ContractTemplate
-	receiver       types.Address
-	closeRemainder types.Address
-	firstValid     uint64
-	lastValid      uint64
-	amount         uint64
-	leaseBase64    string
 }
 
 // MakeDynamicFee contract allows you to create a transaction without
@@ -34,11 +30,11 @@ func MakeDynamicFee(receiver, closeRemainder string, amount, firstValid, lastVal
 	leaseBytes := make([]byte, 32)
 	crypto.RandomBytes(leaseBytes)
 	leaseString := base64.StdEncoding.EncodeToString(leaseBytes)
-	return MakeDynamicFeeWithLease(receiver, closeRemainder, leaseString, amount, firstValid, lastValid)
+	return makeDynamicFeeWithLease(receiver, closeRemainder, leaseString, amount, firstValid, lastValid)
 }
 
-// MakeDynamicFeeWithLease is as MakeDynamicFee, but the caller can specify the lease (using b64 string)
-func MakeDynamicFeeWithLease(receiver, closeRemainder, lease string, amount, firstValid, lastValid uint64) (DynamicFee, error) {
+// makeDynamicFeeWithLease is as MakeDynamicFee, but the caller can specify the lease (using b64 string)
+func makeDynamicFeeWithLease(receiver, closeRemainder, lease string, amount, firstValid, lastValid uint64) (DynamicFee, error) {
 	const referenceProgram = "ASAFAgEHBgUmAyD+vKC7FEpaTqe0OKRoGsgObKEFvLYH/FZTJclWlfaiEyDmmpYeby1feshmB5JlUr6YI17TM2PKiJGLuck4qRW2+SB/g7Flf/H8U7ktwYFIodZd/C1LH6PWdyhK3dIAEm2QaTIEIhIzABAjEhAzAAcxABIQMwAIMQESEDEWIxIQMRAjEhAxBygSEDEJKRIQMQgkEhAxAiUSEDEEIQQSEDEGKhIQ"
 	referenceAsBytes, err := base64.StdEncoding.DecodeString(referenceProgram)
 	if err != nil {
@@ -66,12 +62,6 @@ func MakeDynamicFeeWithLease(receiver, closeRemainder, lease string, amount, fir
 			address: address.String(),
 			program: injectedBytes,
 		},
-		receiver:       receiverAddr,
-		closeRemainder: closeRemainderAddr,
-		firstValid:     firstValid,
-		lastValid:      lastValid,
-		amount:         amount,
-		leaseBase64:    lease,
 	}
 	return dynamicFee, err
 }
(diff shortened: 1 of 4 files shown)


templates/dynamicFee.go
```diff
@@ -0,0 +1,142 @@
...
+//  - maxFee: maximum fee used by the withdrawal transaction
+func MakeDynamicFee(receiver, closeRemainder string, amount, firstValid, lastValid uint64) (DynamicFee, error) {
+	leaseBytes := make([]byte, 32)
+	crypto.RandomBytes(leaseBytes)
+	leaseString := base64.StdEncoding.EncodeToString(leaseBytes)
+	return MakeDynamicFeeWithLease(receiver, closeRemainder, leaseString, amount, firstValid, lastValid)
+}
+
+// MakeDynamicFeeWithLease is as MakeDynamicFee, but the caller can specify the lease (using b64 string)
+func MakeDynamicFeeWithLease(receiver, closeRemainder, lease string, amount, firstValid, lastValid uint64) (DynamicFee, error) {
```
rotemh: The caller can't specify her/is lease. It musts match the original one.

templates/dynamicFee.go
```diff
@@ -0,0 +1,142 @@
...
+		return nil, err
+	}
+	return append(stx1Bytes, stx2Bytes...), nil
+}
+
+// SignDynamicFee returns the main transaction and signed logic needed to complete the
+// transfer. These should be sent to the fee payer, who can use
+// GetDynamicFeeTransactions() to update fields and create the auxiliary
+// transaction.
+func (c DynamicFee) SignDynamicFee(privateKey ed25519.PrivateKey, genesisHash []byte) (txn types.Transaction, lsig types.LogicSig, err error) {
```
rotemh: This one seems problematic. Do you expect the user to re-instantiate the contract? Otherwise, all the `c.element` won't exist. I'd expect this function to make a good use of `GetDynamicFeeTxns` output

templates/dynamicFee.go
EvanJRichard: It's mainly for testing, so that a contract with the specified lease from the test vector can be created and the `golden` checked.

templates/dynamicFee.go
EvanJRichard: I did expect the user to re-instantiate the contract - I'll revisit the design on Confluence and look at the corresponding Python PR to see if there's a different way to do all this.

templates/dynamicFee.go
winder: You should make the method private

templates/dynamicFee.go
EvanJRichard: Changing this to accept `contract []byte`, like Periodic Payment.

templates/templates_test.go
```diff
@@ -39,3 +39,35 @@ func TestHTLC(t *testing.T) {
...
+	require.Equal(t, goldenAddress, c.GetAddress())
+	goldenStxns := "gqRsc2lngqFsxLEBIAUCAYgnuWC6YCYDIP68oLsUSlpOp7Q4pGgayA5soQW8tgf8VlMlyVaV9qITIOaalh5vLV96yGYHkmVSvpgjXtMzY8qIkYu5yTipFbb5IH+DsWV/8fxTuS3BgUih1l38LUsfo9Z3KErd0gASbZBpMgQiEjMAECMSEDMABzEAEhAzAAgxARIQMRYjEhAxECMSEDEHKBIQMQkpEhAxCCQSEDECJRIQMQQhBBIQMQYqEhCjc2lnxEAhLNdfdDp9Wbi0YwsEQCpP7TVHbHG7y41F4MoESNW/vL1guS+5Wj4f5V9fmM63/VKTSMFidHOSwm5o+pbV5lYHo3R4boujYW10zROIpWNsb3NlxCDmmpYeby1feshmB5JlUr6YI17TM2PKiJGLuck4qRW2+aNmZWXOAAWq6qJmds0wOaJnaMQgf4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGmjZ3JwxCBRpaRVpA3ImXU4/ENcrzp+jsooLVHC7bF5kCGUK0KORaJsds0wOqJseMQgf4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGmjcmN2xCD+vKC7FEpaTqe0OKRoGsgObKEFvLYH/FZTJclWlfaiE6NzbmTEIIU9h0wnKapwajF0N7K4zy3orGLF+rQ8kLIk/vW6FhPvpHR5cGWjcGF5gqNzaWfEQAilsGaC4M4zfYN5QpvREdHEC0DjI2ZWCXSIwwyUWHg2dzd5gKR2Cqu+iUmiCU1hOTTiOump3PILTgWeG0ZkUAajdHhuiqNhbXTOAAWq6qNmZWXOAATzvqJmds0wOaJnaMQgf4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGmjZ3JwxCBRpaRVpA3ImXU4/ENcrzp+jsooLVHC7bF5kCGUK0KORaJsds0wOqJseMQgf4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGmjcmN2xCCFPYdMJymqcGoxdDeyuM8t6Kxixfq0PJCyJP71uhYT76NzbmTEICuIj6PMWBK0XH0TqQSTWXj6UWxbhN7Y9jUpXyQ1xxxGpHR5cGWjcGF5"
```
winder: Could you split this into 2 for testing each txn separately?

EvanJRichard: `GetDynamicFeeTransactions` returns one concatenated byte array so that the txn group suitable for passing immediately to `sendRaw`, so I'm not sure it's practical to do that - I'll look, though, you're right, if we can do that it seems like a win for clarity

templates/templates_test.go
```diff
@@ -39,3 +39,35 @@ func TestHTLC(t *testing.T) {
...
+	c, err := makeDynamicFeeWithLease(receiver, closeRemainder, artificialLease, amount, firstValid, lastValid)
+	require.NoError(t, err)
+	goldenGenesisHash := "f4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGk="
+	genesisBytes, err := base64.StdEncoding.DecodeString(goldenGenesisHash)
+	require.NoError(t, err)
+	contractBytes := c.GetProgram()
+	privateKeyOneB64 := "cv8E0Ln24FSkwDgGeuXKStOTGcze5u8yldpXxgrBxumFPYdMJymqcGoxdDeyuM8t6Kxixfq0PJCyJP71uhYT7w=="
+	privateKeyOne, err := base64.StdEncoding.DecodeString(privateKeyOneB64)
+	require.NoError(t, err)
+	txn, lsig, err := SignDynamicFee(contractBytes, privateKeyOne, genesisBytes)
```
winder: Need to make sure we test the Base64 representation of the txn & lsig objects, since that is what will be sent between users.

These are what I generated with the Java SDK:
```
		String goldenTxn = "iaNhbXTNE4ilY2xvc2XEIOaalh5vLV96yGYHkmVSvpgjXtMzY8qIkYu5yTipFbb5omZ2zTA5omdoxCB/g7Flf/H8U7ktwYFIodZd/C1LH6PWdyhK3dIAEm2QaaJsds0wOqJseMQgf4OxZX/x/FO5LcGBSKHWXfwtSx+j1ncoSt3SABJtkGmjcmN2xCD+vKC7FEpaTqe0OKRoGsgObKEFvLYH/FZTJclWlfaiE6NzbmTEIIU9h0wnKapwajF0N7K4zy3orGLF+rQ8kLIk/vW6FhPvpHR5cGWjcGF5";
		String goldenLsig = "gqFsxLEBIAUCAYgnuWC6YCYDIP68oLsUSlpOp7Q4pGgayA5soQW8tgf8VlMlyVaV9qITIOaalh5vLV96yGYHkmVSvpgjXtMzY8qIkYu5yTipFbb5IH+DsWV/8fxTuS3BgUih1l38LUsfo9Z3KErd0gASbZBpMgQiEjMAECMSEDMABzEAEhAzAAgxARIQMRYjEhAxECMSEDEHKBIQMQkpEhAxCCQSEDECJRIQMQQhBBIQMQYqEhCjc2lnxEAhLNdfdDp9Wbi0YwsEQCpP7TVHbHG7y41F4MoESNW/vL1guS+5Wj4f5V9fmM63/VKTSMFidHOSwm5o+pbV5lYH";
```

templates/templates_test.go
winder: I think you're right. They should be tested together.

PR #115, opened 2020-02-07
Fix HTLC templates_test goldens

EvanJRichard:


winder:
You should add in the missing calidation as well: #117

PR #121, opened 2020-02-25
Get raw block

algobolson:
Client code to request raw msgpack bytes for a block from algod.
Also export functions to calculate a txid from a Transaction.
Add .IsZero() for Address and AssetParams{}


crypto/crypto.go
```diff
@@ -79,6 +79,20 @@ func txIDFromTransaction(tx types.Transaction) (txid string) {
 	return
 }
 
+// TransactionID is the unique identifier for a Transaction in progress
```
EvanJRichard: Does this duplicate `txIDFromTransaction` above? / could the usages of `txIDFromTransaction` be reworked to use this new public helper?

client/algod/wrappers.go
```diff
@@ -151,6 +153,34 @@ func (client Client) Block(round uint64, headers ...*Header) (response models.Bl
...
+	var resp *http.Response
+	request := struct {
+		Raw string `url:"raw"`
+	}{Raw: "1"}
+	resp, err = client.submitFormRaw(fmt.Sprintf("/block/%d", round), request, "GET", false, headers)
+	if err != nil {
+		return
+	}
+	defer resp.Body.Close()
+	return responseReadAll(resp, 10000000)
```
EvanJRichard: Would naming `10000000` with a constant be more readable here, or would that be wasted breath?

types/address.go
```diff
@@ -26,6 +26,13 @@ func (a Address) String() string {
 	return base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(checksumAddress)
 }
 
+var zeroAddr = [hashLenBytes]byte{}
```
EvanJRichard: There's an issue open to make `ZeroAddr` a constant available to devs, so exporting this constant might knock that issue out too.

PR #122, opened 2020-03-03
use SuggestedParams rather than separate common-to-all-txns args

EvanJRichard:
## Summary
Many transaction-related functions take a validity period, a fee, whether that fee is flat or per-byte, and so on. This PR proposes to add a struct `SuggestedParams`, returned when `algodClient.SuggestedParams` is called (or constructed manually by a user), that is instead passed to these functions.

### Testing
Unit tests pass. Cucumber testing is known-broken. 

### See also 
https://github.com/algorand/algorand-sdk-testing/pull/48


client/algod/wrappers.go
```diff
@@ -141,6 +142,21 @@ func (client Client) SuggestedParams(headers ...*Header) (response models.Transa
 	return
 }
 
+// BuildSuggestedParams gets the suggested transaction parameters and
```
winder: I think you should add to `SuggestedParams` instead of creating a new type. `LastRound` might be a little strange, but perhaps you can have `LastRound` and `LastRoundValid`? I guess you could also add `FirstRoundValid` and initialize it to the same value as `LastRound`.

templates/templates_test.go
```diff
@@ -2,6 +2,7 @@ package templates
 
 import (
 	"encoding/base64"
+	"github.com/algorand/go-algorand-sdk/types"
```
winder: fix import grouping

future/transaction.go
```diff
@@ -0,0 +1,530 @@
...
+		eSize, err := transaction.EstimateSize(tx)
+		if err != nil {
+			return types.Transaction{}, err
+		}
+		tx.Fee = types.MicroAlgos(eSize * uint64(params.Fee))
+	}
+
+	if tx.Fee < MinTxnFee {
+		tx.Fee = MinTxnFee
+	}
```
winder: There was an idea to update the fee while signing the transaction, did the python SDK do that?

client/algod/wrappers.go
EvanJRichard: @winder I was thinking that leaving `models.TransactionParams` as exactly the struct returned from `algod`, and having a separate struct for the `Transaction` helpers, would make it clearer what `algod` is returning. What you propose is closer to the design document and python implementation though. What would the `json:` comments look like for these extra fields, do you know if there's an `omit` type flag?

future/transaction.go
EvanJRichard: @winder As far as I can tell, no, changing signing to be the step that updates the fee isn't in the scope of this ticket.

client/algod/wrappers.go
EvanJRichard: In the same vein, another goal was to keep the `SuggestedParams()` function returning the same object

future/transaction.go
winder: that seems fine

client/algod/wrappers.go
winder: I guess whatever is most idiomatic is fine. You're right that the field names aren't perfect.

Keeping the original object and changing the constructor to something like this might be a better overall approach in the future:

https://algorand.atlassian.net/wiki/spaces/FDE/pages/491716666/SuggestedParams+Thoughts#Functional-Options

EvanJRichard: @winder I think we'll move towards that with the indexer work (ex. an options struct using `url:` type options like `client.PendingTransaction` currently does, but I digress) for the transaction builders I wanted to hew close to the previous implementation and docs/keep the change smaller. I think this will be cleaned up more with the incoming indexer work, though.

PR #126, opened 2020-03-17
AlgodClientv2 and IndexerClient

EvanJRichard:
# Summary
This PR proposes to add two new clients, `v2client.Client` for algod and `indexer.Client` for the indexer, in a new directory `v2client`. They both wrap the new `common.Client` that can be found in the same directory. 

## Testing
This PR proposes to add unit tests through cucumber, see PR https://github.com/algorand/algorand-sdk-testing/pull/54. Mock json data has not yet been generated, but test steps are available for review.


v2client/algod/v2wrappers.go
```diff
@@ -0,0 +1,118 @@
...
+		// TODO built result from response
+	} else if params.Format == "msgpack" {
+		var response models.RawBlockMsgpack
+		err = client.get(ctx, &response, fmt.Sprintf("/block/%d", round), params, headers)
+		// TODO built result from response
+	} else {
+		err = fmt.Errorf("unrecognized format %s, valid formats are json or msgpack", params.Format)
+	}
+	return
+}
```
EvanJRichard: @winder the schema only says `RawBlock` might come back, but it might come back in two different formats. Is this strategy of attempting to decode into two different structs appropriate?

v2client/algod/v2wrappers.go
```diff
@@ -0,0 +1,118 @@
...
+// TODO ejr received peer feedback to have Block return types.Block not models.Block?
+func (client Client) Shutdown(ctx context.Context, timeout models.ShutdownParams, headers ...*common.Header) error {
+	// TODO EJR how to handle private security token
+	return client.post(ctx, nil, "/shutdown", timeout, headers)
+}
+
+func (client Client) RegisterParticipationKeys(ctx context.Context, account string, params models.RegisterParticipationKeysAccountIdParams, headers ...*common.Header) error {
+	// TODO EJR how to handle private security token
+	return client.post(ctx, nil, fmt.Sprintf("/register-participation-keys/%s", account), nil, headers)
+}
```
EvanJRichard: @winder These are marked Private in the specification. Do they need any extra arguments or anything as a result, or can they be handled just like any other endpoint?

test/helpers.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+func loadMockJson(filename string) ([]byte, error) {
+	// TODO EJR where should the mockjsons be stored? or should the feature file just hold the full path somehow
+	return nil, nil
+}
```
EvanJRichard: @winder I wonder where we should store the mock jsons to pass in to the test, and how best to generate the mock jsons

v2client/common/models/v2models.go
```diff
@@ -0,0 +1,1009 @@
...
+
+	// Round at which the results were computed.
+	CurrentRound uint64 `json:"current-round"`
+}
+
+// AssetBalancesResponse defines model for AssetBalancesResponse.
+type AssetBalancesResponse struct {
+
+	// A simplified version of AssetHolding
+	Balances []MiniAssetHolding `json:"balances"`
```
EvanJRichard: @winder the spec specifies this as a single MiniAssetHolding but it looks like it should come back as an array, so I changed that.

v2client/common/common.go
```diff
@@ -0,0 +1,211 @@
...
+var unversionedPaths = map[string]bool{
+	"/versions": true,
+	"/health":   true,
+}
+
+// rawRequestPaths is a set of paths where the body should not be urlencoded
+var rawRequestPaths = map[string]bool{
+	"/transactions": true,
+}
+
```
EvanJRichard: @winder I wasn't able to understand from the spec, should `transactions` still be a rawRequestPath? See L128 below

v2client/algod/v2wrappers.go
rotemh: No, it depends on the user not the server. The users should specify what format they want it. See `format` query parameter in the spec.

v2client/algod/v2wrappers.go
```diff
@@ -0,0 +1,116 @@
...
+		// TODO built result from response
+	} else if params.Format == "msgpack" {
+		err = client.get(ctx, &result, fmt.Sprintf("/transactions/pending/%s", txid), params, headers)
+	} else {
+		err = fmt.Errorf("unrecognized format %s, valid formats are json or msgpack", params.Format)
+	}
+	return
+}
+
+func (client Client) SendRawTransaction(ctx context.Context, txBytes []byte, headers ...*common.Header) (txid string, err error) {
```
rotemh: Please add x-binary type header

v2client/indexer/indexerv2wrappers.go
```diff
@@ -0,0 +1,88 @@
+package indexer
+
+import (
+	"context"
+	"fmt"
+	"github.com/algorand/go-algorand-sdk/v2client/common"
```
rotemh: Consider making the structure to `go-algorand-sdk/client/v2/common`

v2client/algod/v2wrappers.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (client Client) Shutdown(ctx context.Context, timeout models.ShutdownParams, headers ...*common.Header) error {
+	return client.post(ctx, nil, "/shutdown", timeout, headers)
+}
+
+func (client Client) RegisterParticipationKeys(ctx context.Context, account string, params models.RegisterParticipationKeysAccountIdParams, headers ...*common.Header) error {
+	return client.post(ctx, nil, fmt.Sprintf("/register-participation-keys/%s", account), nil, headers)
+}
+
+func (client Client) PendingTransactionInformation(ctx context.Context, txid string, params models.GetPendingTransactionsParams, headers ...*common.Header) (result types.Transaction, err error) {
```
rotemh: it returns `SignedTransaction` one and the response actually contains more information than that. @winder can elaborate.

EvanJRichard: @rotemh @winder Do I follow correctly: 
- this method should return `types.SignedTransaction`
- the actual client will internally receive a `models.Transaction` from the server - https://github.com/algorand/indexer/blob/master/api/generated/types.go#L307 - but that has the information in it needed to rebuild the `types.SignedTransaction`.

rotemh: if you ask for the `format=msgpack`, it should be already in `types.SignedTransaction` and you can just unmarshall it. The node will return it as it has it.

test/helpers.go
winder: put them in the `algorand-sdk-testing` repo. The next logical step for that repo is probably embedding it as a sub-repo of each SDK. That way we can assume the files exist in a relative directory.

v2client/algod/v2wrappers.go
winder: Evan take a look here:
https://algo-indexer-v2.docs.stoplight.io/reference/v2-transactions-pending-txid/pendingtransactioninformation?__build_id=58434

You can see additional data included (pool error, some round information, etc).

The SDKs are responsible for converting the `txn` field into a `SignedTransaction`

v2client/algod/v2wrappers.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (client Client) SendRawTransaction(ctx context.Context, txBytes []byte, headers ...*common.Header) (txid string, err error) {
+	var response models.TxId
+	err = client.post(ctx, &response, "/transactions", nil, headers)
+	txid = string(response)
+	return
+}
+
+func (client Client) PendingTransactionsByAddress(ctx context.Context, account string, params models.GetPendingTransactionsByAddressParams, headers ...*common.Header) (result []types.Transaction, err error) {
```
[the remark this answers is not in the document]
winder: Here is the endpoint definition. Again as Rotem said, we want to use the existing `SignedTransaction` for this one.

Also notice that it is wrapped in an object which also includes the total size of the pending transaction pool.

https://algo-indexer-v2.docs.stoplight.io/reference/v2-transactions-pending/getpendingtransactions?__build_id=58434

v2client/algod/v2wrappers.go
```diff
@@ -0,0 +1,116 @@
...
+	if params.Format == "json" {
+		var response []models.Transaction
+		err = client.get(ctx, &response, fmt.Sprintf("/accounts/%s/transactions/pending", account), params, headers)
+		// TODO built result from response
+	} else if params.Format == "msgpack" {
+		err = client.get(ctx, &result, fmt.Sprintf("/accounts/%s/transactions/pending", account), params, headers)
+	} else {
+		err = fmt.Errorf("unrecognized format %s, valid formats are json or msgpack", params.Format)
+	}
+	return
```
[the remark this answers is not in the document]
winder: I mentioned this in slack, putting it here for completeness. I think that `params.Format` should not be something that the user can see. I expect that `format=json` would be used by 3rd party developers who don't want to use our SDK, and don't want to figure out how to decode msgpack.

A simple solution might be to change it to `params.format` and set it to `msgpack` inside these functions.

client/v2/common/common.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/algorand/go-algorand-sdk/encoding/msgpack"
+	"github.com/google/go-querystring/query"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+)
+
+const (
+	authHeader = "X-Algo-API-Token"
```
winder, line 17: need one of these for indexer also: `X-Indexer-API-Token`

client/v2/algod/accountInformation.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/algorand/go-algorand-sdk/client/v2/common"
+	"github.com/algorand/go-algorand-sdk/client/v2/common/models"
+)
+
+type AccountInformation struct {
+	c       *Client
+	account string
+}
+
+func (s *AccountInformation) Do(ctx context.Context, headers ...*common.Header) (result models.Account, err error) {
```
winder, line 15: I'm a little confused by the `common.Header` pattern. These endpoints have different parameters, how do you deal with the fact that `/v2/accounts` doesn't support filtering by a txid but `/v2/transactions` does?

winder: The server will return an error, so we're not dead in the water, but if it was organized differently those sorts of errors could be caught by the compiler.

client/v2/algod/algod.go
```diff
@@ -0,0 +1,88 @@
...
+
+type Client common.Client
+
+// get performs a GET request to the specific path against the server, assumes JSON response
+func (c *Client) get(ctx context.Context, response interface{}, path string, request interface{}, headers []*common.Header) error {
+	return (*common.Client)(c).Get(ctx, response, path, request, headers)
+}
+
+// getMsgpack performs a GET request to the specific path against the server, assumes msgpack response
+func (c *Client) getMsgpack(ctx context.Context, response interface{}, path string, request interface{}, headers []*common.Header) error {
```
winder, line 16: Note that if there was an error, the error response will be JSON. You can check that the status code is 200 to decide whether to treat it as a messagepack object.

client/v2/algod/accountInformation.go
winder: nevermind! I mistook headers and parameters

client/v2/algod/block.go
```diff
@@ -0,0 +1,34 @@
...
+
+type Block struct {
+	c     *Client
+	round uint64
+	p     models.GetBlockParams
+}
+
+type generatedBlockResponse struct {
+	// Block header data.
+	Block types.Block `json:"block"`
```
winder, line 19: Is `types.Block` the block object from the indexer endpoint? If so, that isn't the right type. This should also be `map[string]interface{}`

client/v2/algod/pendingTransactionsByAddress.go
```diff
@@ -0,0 +1,29 @@
...
+
+func (s *PendingTransactionInformationByAddress) Max(max uint64) *PendingTransactionInformationByAddress {
+	s.p.Max = max
+	return s
+}
+
+func (s *PendingTransactionInformationByAddress) Do(ctx context.Context, headers ...*common.Header) (total uint64, topTransactions []types.SignedTxn, err error) {
+	s.p.Format = "msgpack"
+	response := models.PendingTransactionsResponse{}
+	err = s.c.getMsgpack(ctx, &response, fmt.Sprintf("/v2/accounts/%s/transactions/pending", s.address), s.p, headers)
```
winder, line 25: Does err include the status code + error message? Maybe we should make a `ResponseError` implementation of `error` so that users can easily access that metadata if they want to.

client/v2/indexer/lookupAccountTransactions.go
```diff
@@ -0,0 +1,100 @@
...
+	s.p.AssetId = index
+	return s
+}
+
+func (s *LookupAccountTransactions) Limit(limit uint64) *LookupAccountTransactions {
+	s.p.Limit = limit
+	return s
+}
+
+func (s *LookupAccountTransactions) BeforeTime(before string) *LookupAccountTransactions {
```
winder, line 67: nit: would be nice to have a date type here, building RFC3339 strings can be annoying.

client/v2/indexer/lookupAccountTransactions.go
```diff
@@ -0,0 +1,100 @@
...
+	s.p.CurrencyGreaterThan = greaterThan
+	return s
+}
+
+func (s *LookupAccountTransactions) CurrencyLessThan(lessThan uint64) *LookupAccountTransactions {
+	s.p.CurrencyLessThan = lessThan
+	return s
+}
+
+func (s *LookupAccountTransactions) AddressRole(role string) *LookupAccountTransactions {
```
winder, line 87: nit: this could be an enum

client/v2/indexer/lookupAccountTransactions.go
```diff
@@ -0,0 +1,100 @@
...
+	s.p.NotePrefix = base64.StdEncoding.EncodeToString(prefix)
+	return s
+}
+
+func (s *LookupAccountTransactions) TxType(txtype string) *LookupAccountTransactions {
+	s.p.TxType = txtype
+	return s
+}
+
+func (s *LookupAccountTransactions) SigType(sigtype string) *LookupAccountTransactions {
```
winder, line 32: nit: this could be an enum

client/v2/indexer/lookupAccountTransactions.go
```diff
@@ -0,0 +1,100 @@
...
+	s.p.NextToken = nextToken
+	return s
+}
+
+func (s *LookupAccountTransactions) NotePrefix(prefix []byte) *LookupAccountTransactions {
+	s.p.NotePrefix = base64.StdEncoding.EncodeToString(prefix)
+	return s
+}
+
+func (s *LookupAccountTransactions) TxType(txtype string) *LookupAccountTransactions {
```
winder, line 27: nit: this could be an enum

client/v2/indexer/lookupAssetBalances.go
```diff
@@ -0,0 +1,50 @@
...
+	s.p.NextToken = nextToken
+	return s
+}
+
+func (s *LookupAssetBalances) Limit(lim uint64) *LookupAssetBalances {
+	s.p.Limit = lim
+	return s
+}
+
+func (s *LookupAssetBalances) AfterAddress(after string) *LookupAssetBalances {
```
winder, line 27: This no longer exists, replaced by NextToken IIRC

client/v2/indexer/lookupAssetTransactions.go
```diff
@@ -0,0 +1,100 @@
...
+	s.p.MinRound = min
+	return s
+}
+
+func (s *LookupAssetTransactions) MaxRound(max uint64) *LookupAssetTransactions {
+	s.p.MaxRound = max
+	return s
+}
+
+func (s *LookupAssetTransactions) Address(address string) *LookupAssetTransactions {
```
winder, line 57: nit: this could be our address type, that would ensure that the address was valid before making the request.

client/v2/indexer/searchForAssets.go
```diff
@@ -0,0 +1,55 @@
...
+	s.p.Unit = unit
+	return s
+}
+
+func (s *SearchForAssets) AssetID(id uint64) *SearchForAssets {
+	s.p.AssetId = id
+	return s
+}
+
+func (s *SearchForAssets) AfterAsset(after uint64) *SearchForAssets {
```
winder, line 44: IIRC, this was replaced by `NextToken`

templates/templates_test.go
```diff
@@ -151,3 +153,31 @@ func TestLimitOrder(t *testing.T) {
 	goldenAddress := "LXQWT2XLIVNFS54VTLR63UY5K6AMIEWI7YTVE6LB4RWZDBZKH22ZO3S36I"
 	require.Equal(t, goldenAddress, c.GetAddress())
 }
+
+func TestV2ClientPlayground(t *testing.T) {
```
winder, line 157: are you planning to remove this before committing?

test/algodclientv2_test.go
```diff
@@ -0,0 +1,379 @@
+package main
```
winder, line 1: Did you re-implement the v1 stepfile with the v2 client? nice!

test/algodclientv2_test.go
```diff
@@ -0,0 +1,379 @@
...
+
+func theParsedSendRawTransactionResponseShouldHaveTxid(txid string) error {
+	if txidResponse != txid {
+		return fmt.Errorf("expected txn to have txid %s but actual txid was %s", txidResponse, txid)
+	}
+	return nil
+}
+
+func weMakeAnyPendingTransactionsByAddressCall() error {
+	algodClient, err := algod.MakeClient(mockServer.URL, "")
```
winder, line 141: I'd probably pop this in an init function instead of repeating it in every step definition

types/block.go
```diff
@@ -0,0 +1,194 @@
+package types
```
winder, line 1: :+1:

client/v2/algod/block.go
EvanJRichard: It's not, it's the `codec`-encoded/msgpack-represented block from `go-algorand-sdk/types`

client/v2/algod/pendingTransactionsByAddress.go
EvanJRichard: At the lowest level the response includes status code & error message, then `v2/common/common.go:extractError` uses that to build specific errors such as `BadRequest` `InvalidToken` etc.

client/v2/algod/algod.go
```diff
@@ -0,0 +1,88 @@
...
+	return &PendingTransactionInformationByAddress{c: c, address: address}
+}
+
+func (c *Client) PendingTransactions() *PendingTransactions {
+	return &PendingTransactions{c: c}
+}
+
+func (c *Client) RegisterParticipationKeys(account string) *RegisterParticipationKeys {
+	return &RegisterParticipationKeys{c: c, account: account}
+}
```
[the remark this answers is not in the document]
EvanJRichard: `Shutdown` and `RegisterParticipationKeys` are tested (at least a little) by the all-sdks cucumber unit tests, I could remove them but I would need to remove the tests in `algorand-sdk-testing` as well

client/v2/algod/algod.go
EvanJRichard: :+1: the msgpack decoder is only used in the non-error path

client/v2/algod/algod.go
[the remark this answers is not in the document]
rotemh: That's fine. We should remove them

Issue #110, opened 2020-01-29
Tx POST missing Content-type header


fxgamundi:
The algod `/v1/transactions` POST endpoint expects a binary content, but the `Content-Type: x-binary` header is not set by the Go SDK when making that request (`SendRawTransaction`). Right now the algod node API works without this header, but other APIs might not (e.g. PureStake API).
The Java SDK already does that, so it would be update this SDK to match that behavior.

PR #131, opened 2020-03-27
Content-Type is now set when sending a transaction

fxgamundi:
Modified `SendRawTransaction` so that a `Content-Type: application/x-binary` header is set, only if the user didn't specify a Content-Type header.
This changes are backwards compatible, and address issue #110.
@ejbaran @JasonWeathersby

### DIFF
--- a/client/algod/wrappers.go
+++ b/client/algod/wrappers.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
-	"github.com/algorand/go-algorand-sdk/types"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


client/algod/wrappers.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
```
EvanJRichard: Can you please sort imports alphabetically? With the `go-algorand-sdk` imports coming last. Thank you!

fxgamundi: Sure! Done in latest commit.

PR #134, opened 2020-04-28
copy RekeyTo field into Transaction from go-algorand

algobolson:
comment on ZeroAddress to make lint happy


algobolson:
Thanks Max, missed that, fixed.

PR #146, opened 2020-05-05
Update TEAL langspec.json for TEAL v2

pzbitskiy:


winder:
@EvanJRichard I'm not sure why there's a conflict, could you take a look?

EvanJRichard:
@winder I _think_ we need to force-merge taking `ours` or `theirs`, but I'm not sure which one is correct. @pzbitskiy / @algorandskiy What do you think?

algorandskiy:
Langspec in this PR is already stalled. I'll update the PR later today with the latest and greatest stuff. `logic/bundledSpecInject.go` is a generated file from `langspec.json`.

algorandskiy:
replaced by #163

PR #151, opened 2020-05-11
Add support for rekeying

rotemh:
This PR adds support for rekeying - 
-- Add the ability to add a `rekey-to` field to any transaction
-- `SignTransaction` now checks if the `SK` is different than the sender's corresponding one and adds `AuthAddr` as needed. 
-- Implements the necessary steps for the cucumber unit and integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -997,3 +997,19 @@ params := types.SuggestedParams {
 // if and only if sender == clawback manager for this asset
 txn, err = MakeAssetRevocationTxn(revocationManager, revocationTarget, amount, recipient, note, params, assetIndex);
 ```
+
+## Rekying
+To rekey an account to a new address, simply call the `Rekey` function on any transaction.
+```golang
+...
+tx, err := future.MakePaymentTxn(
+		account.Address.String(), "4MYUHDWHWXAKA5KA7U5PEN646VYUANBFXVJNONBK3TIMHEMWMD4UBOJBI4",
+		amount, nil, "", params
+	)
+// From now, every transaction needs to be sign the SK of the following address
+tx.Rekey("47YPQTIGQEO7T4Y4RWDYWEKV6RTR2UNBQXBABEEGM72ESWDQNCQ52OPASU")
+...
+```
+
+When submitting a transaction from an account that was rekeying, simply use relevant SK. `SignTransaction` will detect 
+that the SK corresponding address is different than the sender's and will set the `AuthAddr` accordingly. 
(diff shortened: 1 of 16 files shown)


crypto/crypto.go
```diff
@@ -37,8 +37,23 @@ func RandomBytes(s []byte) {
...
+	if n != ed25519.PublicKeySize {
+		return [32]byte{}, fmt.Errorf("generated public key has the wrong size, expected %d, got %d", ed25519.PublicKeySize, n)
+	}
+	return a, nil
+}
+
 // SignTransaction accepts a private key and a transaction, and returns the
 // bytes of a signed transaction ready to be broadcasted to the network
+// If the SK's corresponding address is different than the txn sender's, the SK's
+// corresponding address will be assigned as AuthAddr
```
winder, line 56: This is really elegant, I really that you were able to implement this without changing the method signature.

I can also imagine someone grabbing the wrong private key and being annoyed that the transaction failed because the account wasn't actually rekeyed. If there was a separate `SignTransactionAuthAddr`, then this one could return an error if you tried to sign the tx with a key other than the sender.

I'm not sure which approach is best, this is probably fine.

EvanJRichard:
`kmd/requests.go:SignTransactionRequest` needs new field `PublicKey` - in `go-algorand` it is implemented as:
	PublicKey      crypto.PublicKey `json:"public_key"`

Similarly, `client/kmd/wrappers.go` needs a function that uses this new field.

Otherwise, LGTM!

crypto/crypto.go
rotemh: You're right and I also debated that when adding it. The reason I chose this way is (1) to keep it consistent with Adam's implementation and (2) I believe that the majority of cases will be a success so it's better to make it more convenient for the majority rather than the opposite.

README.md
```diff
@@ -996,4 +996,20 @@ params := types.SuggestedParams {
...
+
+## Rekying
+To rekey an account to a new address, simply call the `Rekey` function on any transaction.
+```golang
+...
+tx, err := future.MakePaymentTxn(
+		account.Address.String(), "4MYUHDWHWXAKA5KA7U5PEN646VYUANBFXVJNONBK3TIMHEMWMD4UBOJBI4",
+		amount, nil, "", params
+	)
+// From now, every transaction needs to be sign the SK of the following address
```
algonautshant, line 1009: ... needs to be signed by ...

PR #161, opened 2020-06-12
Indexer client SDK with applications

algonautshant:
The changes in this PR are a merge between the existing code and the auto-generated from the custom Java/Go generator.

The merge involved keeping existing code, not to break published SDK. 

The only change is the removal of AfterAddress from SearchAccounts/SearchForAccounts.


algorandskiy:
I'd appreciate to see commands needed to regenerate the stuff when needed. I guess it is ok to put it here, in PR for now, and then commit - when the generator finalizes.

PR #167, opened 2020-06-24
Applications SDK with Integration/unit tests

algonautshant:
Application tools and integration/unit tests. 
Some of the code is copied over from go-algorand.


test/applications_integration_test.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+		if status.LastRound > stopRound { // last round sent tx is valid
+			return fmt.Errorf("Transaction not accepted.")
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	return nil
+}
+
+func iBuildAnApplicationTransactionWithTheTransientAccountTheCurrentApplicationSuggestedParamsOperationApprovalprogramClearprogramGlobalbytesGlobalintsLocalbytesLocalintsAppargsForeignappsAppaccounts(
```
algorandskiy, line 110: is the name a joke?)

algonautshant: No. Cucumber is.

algorandskiy: Look at this: https://github.com/algorand/go-algorand-sdk/pull/169/files#diff-5390a1ea0a4c329071050d5e6d1b1092R284
you do not need to follow these weird suggestions from godog.

test/applications_unit_test.go
```diff
@@ -0,0 +1,185 @@
...
+		return err
+	}
+	if !bytes.Equal(gold, stx) {
+		return fmt.Errorf("Application signed transaction does not match the golden.")
+	}
+	return nil
+}
+
+func ApplicationsUnitContext(s *godog.Suite) {
+	s.Step(`^mock server recording request paths$`, mockServerRecordingRequestPaths)
```
winder, line 175: Looks like this was added on accident

transaction/transaction.go
```diff
@@ -806,3 +807,117 @@ func byte32FromBase64(in string) (out [32]byte, err error) {
 	copy(out[:], slice)
 	return
 }
+
+var emptySchema = types.StateSchema{}
+
```
winder, line 812: Could you move these above `AssignGroupID` so that they're next to all the other `Make*` functions?

Also, we seem to have `Make*FlatFee` for all of the other make functions, could you add those here too?

winder: Also please follow the same naming convention as the other constructors, remove `Unsigned` from the new methods.

winder: I think there are implementations in a `future` package somewhere that uses the "suggested params" approach, maybe we should only implement the new form of these? Please check with @EvanJRichard about that.

EvanJRichard: +1 to also adding a FlatFee option.

Yes - matching py-sdk, the transaction building functions that use the `SuggestedParams` are in package `future` to avoid breaking backwards compatibility. Because this PR's proposed changes don't break backwards compatibility, they _might_ be okay here in `transaction`. But either way I'd really love to see a version in `future` that accepts the `SuggestedParams` object, my understanding is we want to start nudging user devs towards using `SuggestedParams`.  

I like this new way of building transactions (make a transaction + set a header) but it does break the pattern of how the other transaction building functions work, so if we choose to stick with this method I think we should have extra documentation/examples in the `README`

transaction/transaction.go
```diff
@@ -806,3 +807,117 @@ func byte32FromBase64(in string) (out [32]byte, err error) {
...
+}
+
+func parseTxnForeignApps(foreignApps []uint64) (parsed []types.AppIndex) {
+	for _, aidx := range foreignApps {
+		parsed = append(parsed, types.AppIndex(aidx))
+	}
+	return
+}
+
+func SetHeader(
```
EvanJRichard, line 895: Rephrasing feedback per our call -
- Recommending changing `SetHeader` to use `types.SuggestedParams` for things like the `fee`
- This pattern of using `MakeFoo` paired with `SetHeader` is a new pattern to this sdk, so this likely merits extra documentation and examples in `README.md`
- It is an open design decision whether these new functions should be in `future` or `transaction` - most existing `transaction` functions are deprecated. After talking it out on the phone, I think leaving it in `transaction` makes sense, because this new code does not break backwards compatibility. (`future` was intended only for functions that would break backwards compatibility.)

algonautshant: Here is my understanding of SuggestedParams:
- SuggestedParams overlaps with Header
- SuggestedParams is a convenience utility of the client via BuildSuggestedParams
- SuggestedParams struct is used for copying its values over to Header.

My argument is:
1. SuggestedParams should only be involved when BuildSuggestedParams is used. Without BuildSuggestedParams, it is a redundant copy of Header. 

2. SuggestedParams should not be in competition with Header. It is confusing for a user do decide which one to use. 

For example, in unit tests, where no client exists, there should be no SuggestedParams involved. Only Header and other basic structures should be involved.

Therefore, SuggestedParams should not be part of the application building function interface.

EvanJRichard: :+1: Okay! In that case, please put extra explanation and examples in `README.md` so users and developer relations understand this new way of building transactions. You could also add `SetHeaderFromSuggestedParams`, because user devs seem to really like using `types.SuggestedParams` to fill in header elements, but maybe that's messy/confusing.

winder: Header already has meaning in the context of an HTTP Client, I'm not sure why we would overload that term in this way. I'm also not sure why we would add another convention for setting these values when we already have 2, and plan on introducing builders as well.

algonautshant: > why we would add another convention for setting these values when we already have 2
I think you ask the right question here.
It is because applications transactions already have many fields. Adding on top the basic transaction fields like sender, note... will be too much.

Moreover, the makeTransaction functions do not give a final transaction object. There are examples where other fields are added afterwards. For example: tx.Rekey, tx.Group. 

I will draft another solution, which will replace SetHeader with SetTransactionFields(sg SuggestedParams, sender, note, group, lease, rekeyTo)
This will be a compromise, where the beloved SuggestedParams will be used, without making a freight train out of the makeApplicationTx call.

transaction/transaction.go
```diff
@@ -892,32 +892,42 @@ func parseTxnForeignApps(foreignApps []uint64) (parsed []types.AppIndex) {
 	return
 }
 
-func SetHeader(
-	tx *types.Transaction,
+func SetApplicationTransactionFields(
```
winder, line 895: These parameters should be part of the `MakeApplicationCall*` functions.

transaction/transaction.go
```diff
@@ -806,3 +807,127 @@ func byte32FromBase64(in string) (out [32]byte, err error) {
 	copy(out[:], slice)
 	return
 }
+
+var emptySchema = types.StateSchema{}
+
+// MakeUnsignedAppCreateTx makes a transaction for creating an application
```
winder, line 813: Remove `Unsigned` from these function names.

I'm not sure if we should abbreviate to `App` or use `Application`, but we definitely should not use both like we have now.

transaction/transaction.go
algonautshant: I don't think adding 6 more parameters is a good idea.

transaction/transaction.go
algonautshant: fixed.

transaction/transaction.go
algorotem: It's not, you're right. But, as discussed, the solution will be a builder pattern. We don't want to introduce a 3rd way of building transactions. So right now we should stick to the old (and I agree that not-so-great) convention until we introduce the builder pattern method here.

algonautshant: done.

future/transaction.go
```diff
@@ -3,13 +3,17 @@ package future
...
 	"github.com/algorand/go-algorand-sdk/transaction"
 	"github.com/algorand/go-algorand-sdk/types"
 )
 
 // MinTxnFee is v5 consensus params, in microAlgos
 const MinTxnFee = transaction.MinTxnFee
 
+// NumOfAdditionalBytesAfterSigning is the number of bytes added to a txn after signing it
+const NumOfAdditionalBytesAfterSigning = 75
+
```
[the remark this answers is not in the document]
algonautshant: removed.

future/transaction.go
```diff
@@ -528,3 +529,355 @@ func byte32FromBase64(in string) (out [32]byte, err error) {
...
+//                application. The LocalStateSchema is immutable.
+//
+// - onComplete   This is the faux application type used to distinguish different
+//                application actions. Specifically, OnCompletion specifies what
+//                side effects this transaction will have if it successfully makes
+//                it into a block.
+
+// MakeApplicationCreateTx makes a transaction for creating an application (see above for args desc.)
+func MakeApplicationCreateTx(
+	onComplete types.OnCompletion,
```
winder, line 570: We shouldn't accept `OnCompletion` here, the only valid `OnCompletion` types are `NoOpOC` and `OptInOC`.

I think that the options are either (1) remove the field, or (2) change the field to a boolean `optInSender` (or a better name).

test/applications_integration_test.go
```diff
@@ -0,0 +1,290 @@
...
+	err := waitForTransaction(txid)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func iRememberTheNewApplicationID() error {
+
+	response, err := acl.RawRequest(fmt.Sprintf("/transactions/pending/%s", txid), nil, "GET", false /* encodeJSON */, nil)
```
winder, line 237: I like this utility, I have new feature request for the generator :)

PR #169, opened 2020-06-29
Implement compile and dryrun transport

algorandskiy:
* Add integration and unit tests

### DIFF
--- a/client/v2/algod/dryrun.go
+++ b/client/v2/algod/dryrun.go
@@ -1,0 +1,24 @@
+package algod
+
+import (
+	"context"
+
+	"github.com/algorand/go-algorand-sdk/client/v2/common"
+	"github.com/algorand/go-algorand-sdk/client/v2/common/models"
+)
+
+// TealDryRun /v2/teal/dryrun
+type TealDryRun struct {
+	c      *Client
+	rawobj []byte
+}
+
+// Do performs HTTP request
+func (s *TealDryRun) Do(
+	ctx context.Context,
+	headers ...*common.Header,
+) (response models.DryrunResponse, err error) {
+	err = s.c.post(ctx, &response,
+		"/v2/teal/dryrun", s.rawobj, headers)
+	return
+}
--- a/test/steps_test.go
+++ b/test/steps_test.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"bytes"
+	"context"
 	"crypto/sha256"
 	"encoding/base32"
 	"encoding/base64"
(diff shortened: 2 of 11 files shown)


EvanJRichard:
Looking good, pending resolution of https://github.com/algorand/algorand-sdk-testing/pull/94 ! :+1:

client/v2/algod/dryrun.go
```diff
@@ -0,0 +1,24 @@
...
+	"context"
+
+	"github.com/algorand/go-algorand-sdk/client/v2/common"
+	"github.com/algorand/go-algorand-sdk/client/v2/common/models"
+)
+
+// TealDryRun /v2/teal/dryrun
+type TealDryRun struct {
+	c      *Client
+	rawobj []byte
```
algonautshant, line 13: In the spec, there is a second parameter that can go to the body. Why is it missing here?
            "description": "Transaction (or group) and any accompanying state-simulation data",                                                              
            "name": "jsonobj",                                                                                                                               
            "in": "body",                                                                                                                                    
            "schema": {                                                                                                                                      
              "$ref": "#/definitions/DryrunRequest"                                                                                                          
            }

algorandskiy: correct. this is a way to express in OAS2 the fact the endpoint accepts either blob or json object. I'm open to suggestions how the spec can be changed.

test/steps_test.go
```diff
@@ -112,6 +117,16 @@ var contractTestFixture struct {
 	dynamicFee         templates.DynamicFee
 }
 
+var tealCompleResult struct {
+	status   int
```
winder, line 121: I think that go has error types that you can compare against instead of adding the status to the result here.

The lack of status output in the JS / Python / Go implementations is pretty unfortunate. I don't think there is even a workaround in JS / Python like there is with Go.

algorandskiy: at the end we end up with the same `if/else`
```
	if err == nil {
		tealCompleResult.status = 200
		tealCompleResult.response = result
		return
	}
	if _, ok := err.(commonV2.BadRequest); ok {
		tealCompleResult.status = 400
		tealCompleResult.response.Hash = ""
		tealCompleResult.response.Result = ""
		return nil
	}
```
but not in `tealCompile` but in `tealCheckCompile`. Do not see a reason to move it there.

client/v2/algod/dryrun.go
algorandskiy: Fixed the spec, now it has only one in-body parameter

test/steps_test.go
winder: Sorry, I meant that you don't need to return the status field because it's indirectly returned via the err object.

algorandskiy: here?
```
result, err := aclv2.TealCompile(tealProgram).Do(context.Background())
```

err will be either `nil` or `commonV2 .BadRequest` and still needs to checked against status code from the test spec.

PR #174, opened 2020-07-09
Put application response models into a separate file.

algonautshant:
These models are generated with the generator, and the specs are changing. It is easier to maintain a shorter file where all the models are generated. 
Also, with a fresh start, can make those link compliant. 

This change has no code change. It is simple cut/paste.


algorandskiy:
Just for reference, the commands to re-generate models are (were):
```bash
cd java-algorand-sdk/generator
mvn package
java -cp target/generator-1.0.0-jar-with-dependencies.jar com.algorand.sdkutils.RunAlgodV2Generator `pwd`/go-algorand/daemon/algod/api/algod.oas2.json
```

PR #176, opened 2020-07-20
Response and Path tests

algonautshant:
Added responses and path unit tests. 

Dependency added:
github.com/nsf/jsondiff for comparing json structures

Changes to test helpers:
* Updated mock http responses helper to accept status code
* Implemented EqualJson method to compare json files and not complain about missing default(false) valued fields

Changes to new apis:
* Catchpoint responses added to NodeStatus
* Application fields added to various response models

changes to existing apis:
* Removed HasSyncedSinceStartup bool `json:"has-synced-since-startup from NodeStatus
* SearchForAssets should return models.AssetsResponse otherwise, missing next-token
* Added AuthAddr string `json:"auth-addr,omitempty"` to Transaction in response models.go
* Added IntraRoundOffset uint64 `json:"intra-round-offset,omitempty"` to Transaction in response models.go (is this new?)
* Some existing tests updated due to the SearchForAssets change

Fixed the type of TealPrograms to []byte (was string)

### DIFF
--- a/client/v2/indexer/searchForAssets.go
+++ b/client/v2/indexer/searchForAssets.go
@@ -7,45 +7,54 @@ import (
 	"github.com/algorand/go-algorand-sdk/client/v2/common/models"
 )
 
+// SearchForAssets /v2/assets
+// Search for assets.
 type SearchForAssets struct {
 	c *Client
 	p models.SearchForAssetsParams
 }
 
+// NextToken the next page of results. Use the next token provided by the previous
+// results.
 func (s *SearchForAssets) NextToken(nextToken string) *SearchForAssets {
 	s.p.NextToken = nextToken
 	return s
 }
 
+// Limit maximum number of results to return.
 func (s *SearchForAssets) Limit(lim uint64) *SearchForAssets {
 	s.p.Limit = lim
 	return s
 }
 
+// Creator filter just assets with the given creator address.
 func (s *SearchForAssets) Creator(creator string) *SearchForAssets {
 	s.p.Creator = creator
 	return s
 }
 
+// Name filter just assets with the given name.
 func (s *SearchForAssets) Name(name string) *SearchForAssets {
 	s.p.Name = name
 	return s
 }
 
+// Unit filter just assets with the given unit.
 func (s *SearchForAssets) Unit(unit string) *SearchForAssets {
 	s.p.Unit = unit
 	return s
 }
 
+// AssetID asset ID
 func (s *SearchForAssets) AssetID(id uint64) *SearchForAssets {
 	s.p.AssetId = id
 	return s
 }
 
-func (s *SearchForAssets) Do(ctx context.Context, headers ...*common.Header) (validRound uint64, result []models.Asset, err error) {
-	var response models.AssetsResponse
-	err = s.c.get(ctx, &response, "/v2/assets", s.p, headers)
-	validRound = response.CurrentRound
-	result = response.Assets
+// Do performs HTTP request
+func (s *SearchForAssets) Do(ctx context.Context,
+	headers ...*common.Header) (response models.AssetsResponse, err error) {
+	err = s.c.get(ctx, &response,
+		"/v2/assets", s.p, headers)
 	return
 }
--- a/test/algodclientv2_test.go
+++ b/test/algodclientv2_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"path"
 
 	"github.com/algorand/go-algorand-sdk/client/v2/algod"
 	"github.com/algorand/go-algorand-sdk/client/v2/common/models"
@@ -361,4 +362,9 @@ func parsedDryrunResponseShouldHave(key string, action int) error {
 		return fmt.Errorf("action %d != %d", action, int(gd[0].Value.Action))
 	}
 	return nil
+}
+
+func mockHttpResponsesInLoadedFrom(jsonfiles, directory string) error {
+	fullPath := path.Join("./features/unit/", directory)
+	return mockHttpResponsesInLoadedFromHelper(jsonfiles, fullPath)
 }
--- a/test/algodclientv2_test.go
+++ b/test/algodclientv2_test.go
@@ -365,6 +365,6 @@ func parsedDryrunResponseShouldHave(key string, action int) error {
 }
 
 func mockHttpResponsesInLoadedFrom(jsonfiles, directory string) error {
-	fullPath := path.Join("./features/unit/", directory)
+	fullPath := path.Join("./features/resources/", directory)
 	return mockHttpResponsesInLoadedFromHelper(jsonfiles, fullPath, 0)
 }
--- a/test/algodclientv2_test.go
+++ b/test/algodclientv2_test.go
@@ -366,5 +366,5 @@ func parsedDryrunResponseShouldHave(key string, action int) error {
 
 func mockHttpResponsesInLoadedFrom(jsonfiles, directory string) error {
 	fullPath := path.Join("./features/unit/", directory)
-	return mockHttpResponsesInLoadedFromHelper(jsonfiles, fullPath)
+	return mockHttpResponsesInLoadedFromHelper(jsonfiles, fullPath, 0)
 }
--- a/test/applications_unit_test.go
+++ b/test/applications_unit_test.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"bytes"
+	"context"
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
@@ -10,9 +11,11 @@ import (
 	"github.com/cucumber/godog"
 	"golang.org/x/crypto/ed25519"
 
+	"github.com/algorand/go-algorand-sdk/client/v2/algod"
+	"github.com/algorand/go-algorand-sdk/client/v2/indexer"
 	"github.com/algorand/go-algorand-sdk/crypto"
-	"github.com/algorand/go-algorand-sdk/mnemonic"
 	"github.com/algorand/go-algorand-sdk/future"
+	"github.com/algorand/go-algorand-sdk/mnemonic"
 	"github.com/algorand/go-algorand-sdk/types"
 )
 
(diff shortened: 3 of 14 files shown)


client/v2/indexer/searchForAssets.go
```diff
@@ -7,45 +7,54 @@ import (
...
 }
 
-func (s *SearchForAssets) Do(ctx context.Context, headers ...*common.Header) (validRound uint64, result []models.Asset, err error) {
-	var response models.AssetsResponse
-	err = s.c.get(ctx, &response, "/v2/assets", s.p, headers)
-	validRound = response.CurrentRound
-	result = response.Assets
+// Do performs HTTP request
+func (s *SearchForAssets) Do(ctx context.Context,
+	headers ...*common.Header) (response models.AssetsResponse, err error) {
```
winder, line 56: We probably need to keep the old method signature... dev-rel has been cranking out example code that might need to be updated if we make a change like this.

The the tests could be fixed by construct the `AssetsResponse` object from `validRound` / `[]Asset`.

Let me get back to you on this.

test/algodclientv2_test.go
```diff
@@ -362,3 +363,8 @@ func parsedDryrunResponseShouldHave(key string, action int) error {
 	}
 	return nil
 }
+
+func mockHttpResponsesInLoadedFrom(jsonfiles, directory string) error {
+	fullPath := path.Join("./features/unit/", directory)
```
winder, line 368: are these also in `./features/resources`?

test/helpers.go
```diff
@@ -18,7 +18,7 @@ func loadMockJsons(commaDelimitedFilenames, pathToJsons string) ([][]byte, error
 	jsonFilenames := strings.Split(commaDelimitedFilenames, ",")
 	var bytesArray [][]byte
 	for _, jsonFilename := range jsonFilenames {
-		fullPath := path.Join("./features/unit/", pathToJsons, jsonFilename)
```
winder, line 21: Instead of the other helper, I think this can be hard coded to `./features/resources`

test/applications_unit_test.go
```diff
@@ -165,9 +168,52 @@ func getSuggestedParams(
 	}, err
 }
 
+func weMakeAGetAssetByIDCall(assetID int) error {
+	clt, err := algod.MakeClient(mockServer.URL, "")
+	if err != nil {
+		return err
+	}
+	clt.GetAssetByID(uint64(assetID)).Do(context.Background())
```
winder, line 176: Looks like these tests save the error for additional validation:
`_, globalErrForExamination = algodClient.Block(uint64(blocknum)).Do(context.Background())`

test/responses_unit_test.go
```diff
@@ -0,0 +1,222 @@
...
+		case "searchForAccounts":
+			response, err = indexerC.SearchAccounts().Do(context.Background())
+		case "lookupApplicationByID":
+			response, err = indexerC.LookupApplicationByID(10).Do(context.Background())
+		case "searchForApplications":
+			response, err = indexerC.SearchForApplications().Do(context.Background())
+		case "lookupAssetBalances":
+			response, err = indexerC.LookupAssetBalances(10).Do(context.Background())
+		case "lookupAssetByID":
+			round, something, err = indexerC.LookupAssetByID(10).Do(context.Background())
```
winder, line 71: I see that this is still returning 3 things instead of a response object and an error?

`SearchForAssets` was changed to return a response object. Why are only some of these `Do` functions updated?

test/responses_unit_test.go
```diff
@@ -0,0 +1,222 @@
...
+	return err
+}
+
+type txidresponse struct {
+	TxID string `json:"txId"`
+}
+
+func theParsedResponseShouldEqualTheMockResponse() error {
+	var err error
+
```
winder, line 157: There are error message responses, is there a way to extract them and have that be part of the test verification?

test/responses_unit_test.go
```diff
@@ -0,0 +1,222 @@
...
+	return err
+}
+
+func ResponsesContext(s *godog.Suite) {
+	s.Step(`^mock http responses in "([^"]*)" loaded from "([^"]*)" with status (\d+)\.$`, mockHttpResponsesInLoadedFromWithStatus)
+	s.Step(`^we make any "([^"]*)" call to "([^"]*)"\.$`, weMakeAnyCallTo)
+	s.Step(`^the parsed response should equal the mock response\.$`, theParsedResponseShouldEqualTheMockResponse)
+}
+
+func EqualJson(j1, j2 string) (ans bool, err error) {
```
winder, line 185: This looks really nice. I think I understand whats happening, but could you add a couple comments to explain whats happening?

Doesn't the removed/added case need to be updated to verify that the other field is a zero value?

winder: I think it would be worth updating some of the old tests in order to test this function, let me know what you think.

You already updated the common `mockHttpResponsesInLoadedFrom` function, so I think it's a matter of:
```diff
func weMakeAnyPendingTransactionsInformationCall() error {    
-  algodClient, err := algod.MakeClient(mockServer.URL, "")    
-  if err != nil {    
-    return err    
-  }      
-  _, stxsResponse, globalErrForExamination = algodClient.PendingTransactions().Do(context.Background())    
-  return nil    
+  weMakeAnyCallTo("algod", "PendingTransactions")
+  return nil
}    

func theParsedPendingTransactionsInformationResponseShouldContainAnArrayOfLenAndElementNumberShouldHaveSender(length, idx int, sender string) error {
-  if len(stxsResponse) != length {
-    return fmt.Errorf("expected response length %d but received length %d", length, len(stxsResponse))
-  }
-  if stxsResponse[idx].Txn.Sender.String() != sender {
-    return fmt.Errorf("expected txn %d to have sender %s but real sender was %s", idx, sender, stxsResponse[idx].Txn.Sender.String())
-  }
+  theParsedResponseShouldEqualTheMockResponse()
  return nil
}
```

This might be a bad example, I think `PendingTransactions` is one of the `Format=msgpack` endpoints.

client/v2/indexer/searchForAssets.go
algonautshant: There is a bug in the old signature: it leaves out NextToken (see ApplicationsResponse below). 
The signature will change anyway, and all the usages will need to be updated. 

old signature:
`
func (s *SearchForAssets) Do(ctx context.Context, headers ...*common.Header) (validRound uint64, result []models.Asset, err error) {		`



```

type ApplicationsResponse struct {
	Applications []Application `json:"applications,omitempty"`

	// CurrentRound round at which the results were computed.
	CurrentRound uint64 `json:"current-round,omitempty"`

	// NextToken used for pagination, when making another request provide this token
	// with the next parameter.
	NextToken string `json:"next-token,omitempty"`
}

```

test/algodclientv2_test.go
algonautshant: It is working in Travis. And I do not decide where travis gets them from. So I must be doing something right.

test/helpers.go
algonautshant: If they end up in features/resources.

test/responses_unit_test.go
algonautshant: Let's leave the code cleanup for later, since we are on tight schedule right now. 
I'll add the comments :-)

test/responses_unit_test.go
algonautshant: I am not changing the existing API. 
The only published one changed (AssetsResponse) is because it has a bug anyway.
This is not a bug, just not consistent. And changing it will break the published API. 
If we are to change published API, I'll replace the whole v2 with the generated one :-)

test/applications_unit_test.go
algonautshant: globalErrForExamination plays a role in scenario:
Then expect error string to contain "<err>"
The feature file for responses does not have that scenario.

test/responses_unit_test.go
algonautshant: I'll return the returned error from Do. It should be nil, and returning should do.

test/applications_unit_test.go
winder: Thats with the old response tests format, the new ones should actually check the error response against these:
```
resources/generated_responses/algod_applications_ErrorResponse_0.json
resources/generated_responses/indexer_applications_ErrorResponse_0.json
```

test/algodclientv2_test.go
winder: It's one of the "backwards compatibility" considerations. Once everything has moved to `resources` (hopefully after responses v2 are implemented for all SDKs) we'll delete the duplicate response files in `unit`.

client/v2/indexer/searchForAssets.go
winder: Oh, nice catch. Thanks for the explanation

test/responses_unit_test.go
algonautshant: > Doesn't the removed/added case need to be updated to verify that the other field is a zero value?

We will add it as we get a case for it. Right now, non is observed.

test/algodclientv2_test.go
winder, line 368: Could you change this to `features/resources`? The files in `unit` are also in `resources`, I'd like to delete `unit`. For example:
```
$ find . |grep pendingTransactionsByAddr.base64
./test/features/unit/v2algodclient_responsejsons/pendingTransactionsByAddr.base64
./test/features/resources/v2algodclient_responsejsons/pendingTransactionsByAddr.base64
```

test/responses_unit_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+func theParsedResponseShouldEqualTheMockResponse() error {
+	var err error
+	var responseJson string
+
+	if expectedStatus != 200 {
+		responseJson = response.(error).Error()
+		// The error message is not a well formed Json.
+		// Verify the expected status code, and remove the json corrupting message
+		extraTxt := fmt.Sprintf("HTTP %d Internal Server Error:", expectedStatus)
```
winder, line 143: Could you implement this in a way that supports other error codes?

For example `responseJson = strings.Split(responseJson, ':')[1]`
or `responseJson = responseJson[strings.Index(responseJson, "{"):]`

test/algodclientv2_test.go
algonautshant: done. I guess I did not completely understand you before. 
Thanks.

test/responses_unit_test.go
algonautshant: done

PR #178, opened 2020-07-30
Implementation of @indexer.applications tests

algonautshant:
In this PR:
* Implementation of @indexer.applications tests.
* Addition of missing application fields in models.
* Implementation of a tool to sort elements inside json arrays.
* The sort function ignores the elements with default values.

### DIFF
--- a/test/indexer_integration_test.go
+++ b/test/indexer_integration_test.go
@@ -592,33 +592,13 @@ func iUseToSearchForTransactionsWithAppIdAndToken(
 	indexer, limit int, notePrefix, txType, sigType, txId string, round,
 	minRound, maxRound, assetId int, beforeTime, afterTime string,
 	currencyGt, currencyLt int, address, addressRole, excludeCloseTo string, appId int, token string) error {
-	ic := indexerClients[indexer]
+
 	var err error
-	notePrefixBytes, err := base64.StdEncoding.DecodeString(notePrefix)
-	if err != nil {
-		return err
-	}
-	excludeBool, err := strconv.ParseBool(excludeCloseTo)
-	if err != nil {
-		excludeBool = false
-	}
+	ic := indexerClients[indexer]
 	response, err = ic.SearchForTransactions().
 		ApplicationId(uint64(appId)).
 		Limit(uint64(limit)).
-		NotePrefix(notePrefixBytes).
-		TxType(txType).SigType(sigType).
-		TXID(txid).Round(uint64(round)).
-		MinRound(uint64(minRound)).
-		MaxRound(uint64(maxRound)).
-		AssetID(uint64(assetId)).
-		BeforeTimeString(beforeTime).
-		AfterTimeString(afterTime).
-		CurrencyGreaterThan(uint64(currencyGt)).
-		CurrencyLessThan(uint64(currencyLt)).
-		AddressString(address).
-		AddressRole(addressRole).
-		ExcludeCloseTo(excludeBool).
-		NextToken(token).Do(context.Background())
+		Do(context.Background())
 	return err
 }
 
--- a/test/indexer_integration_test.go
+++ b/test/indexer_integration_test.go
@@ -4,13 +4,17 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
 	"strconv"
 	"time"
 
+	"github.com/cucumber/godog"
+
 	"github.com/algorand/go-algorand-sdk/client/v2/common/models"
 	"github.com/algorand/go-algorand-sdk/client/v2/indexer"
-
-	"github.com/cucumber/godog"
+	"github.com/algorand/go-algorand-sdk/encoding/json"
 )
 
 func IndexerIntegrationTestContext(s *godog.Suite) {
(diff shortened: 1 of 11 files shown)


test/utilities.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// SortJsonArrays sorts the elements inside json arrays.
+// Some new line organization may be broken.
+// e.g. ...[  {a: "a3", b: "b3"},   {a: "a1", b: "b1"},    {a: "a2", b: "b2"}]...
+// will be:
+//      ...[{a: "a1", b: "b1"},{a: "a2", b: "b2"},{a: "a3", b: "b3"}]...
+// Note: any characters between [ and {, as well as between elements will be lost: ("...},xxx{..." "xxx" is lost ) 
+// This function will recursively sort arrays inside arrays as well.
+func SortJsonArrays(json string) (sorted string, err error) {
```
winder, line 104: Do the tests pass if you only sort the maps `{}`, and not the arrays `[]`?

test/utilities.go
```diff
@@ -0,0 +1,196 @@
...
+	"strings"
+
+	"github.com/nsf/jsondiff"
+)
+
+// EqualJson compares two json strings.
+// returns true if considered equal, false otherwise.
+// The error returns the difference.
+// For reference: j1 is the baseline, j2 is the test
+func EqualJson(j1, j2 string) (ans bool, err error) {
```
winder, line 16: The file should already be sorted, does this test pass if you only sort the actual object? (j2)

test/utilities.go
```diff
@@ -0,0 +1,196 @@
...
+	// Check the difference, and decide if it is still accepted as "equal"
+	// Scan line by line, and examine the difference
+	scanner := bufio.NewScanner(strings.NewReader(str))
+	for scanner.Scan() {
+		line := scanner.Text()
+		// If baseline has the property (j1), but the test (j2) has it removed
+		if strings.Contains(line, "___REMED___") {
+			// If the value of the property is false, then accept it
+			// The default value is false. Missing property is considered as false
+			propValue := strings.SplitN(line, ": ", 2)
```
winder, line 54: Could you give an example of what line would be? Specifically the `len(propValue) == 1` case.

test/utilities.go
```diff
@@ -0,0 +1,196 @@
...
+			return -1, fmt.Errorf("Could not find the matching %c", cchar)
+		}
+	}
+	return -1, fmt.Errorf("Could not find the matching %c", cchar)	
+	//	return -1, fmt.Errorf("Could not find the matching %c", cchar)
+}
+
+// sorts the elements inside the array
+// if a nested arrray is detected, it will first sort the elements inside the nested array
+func sortElements(json *string, start, end int, ochar, cchar rune) (closeIndex int, err error) {
```
winder, line 160: Were you able to verify exactly where the unordered elements came from? We verified with `watch` that Indexer sorts it's response, and the documentation of json/encode claims that it also sorts: https://golang.org/src/encoding/json/encode.go?#L136

test/utilities.go
algonautshant: I only sort the arrays.

test/utilities.go
algonautshant: I did not check. But it is irrelevant. The sorted order is not valid if there are default value fields which are not in the actual object. The actual order sorted will have a different order.

test/utilities.go
algonautshant: See my comment above, regarding the default value issue. 
Moreover, both the response and the object jsons do not have proper ordering of the elements inside arrays. This ordering depends on the values, and not just json keys.

test/utilities.go
algonautshant: ___REMED___},

winder: So `{ "field":"value"` compared to `{"field":"value"}` would be considered equal?

test/indexer_integration_test.go
```diff
@@ -592,33 +592,13 @@ func iUseToSearchForTransactionsWithAppIdAndToken(
...
-	notePrefixBytes, err := base64.StdEncoding.DecodeString(notePrefix)
-	if err != nil {
-		return err
-	}
-	excludeBool, err := strconv.ParseBool(excludeCloseTo)
-	if err != nil {
-		excludeBool = false
-	}
+	ic := indexerClients[indexer]
 	response, err = ic.SearchForTransactions().
```
winder, line 598: Why did you need to remove all of the parameters?

algonautshant: The test was failing when running with other feature tags. 
The cause of the failure, I noticed, was that it was setting a value to txid.

    When I use <indexer> to search for transactions with <limit>, "", "", "", "", 0, 0, 0, 0, "", "", 0, 0, "", "", "", <application-id> and token "" # indexer_integration_test.go:592 -> iUseToSearchForTransactionsWithAppIdAndToken
    Then the parsed response should equal "<jsonfile>".                                                                                               # indexer_integration_test.go:566 -> theParsedResponseShouldEqual

    Examples:
      | indexer | limit | application-id | jsonfile                                                             |
      | 2       | 0     | 70             | v2indexerclient_responsejsons/indexer_v2_tx_search_app_70.json       |
      | 2       | 3     | 70             | v2indexerclient_responsejsons/indexer_v2_tx_search_app_70_lim_3.json |

 http://localhost:59998/v2/transactions?application-id=70&txid=THVWAL7D2N2C4VLYKOWN2K6KG3O3ZDSA356AP3JMQMMGGUYCWKLQ
http://localhost:59998/v2/transactions?application-id=70&limit=3&txid=THVWAL7D2N2C4VLYKOWN2K6KG3O3ZDSA356AP3JMQMMGGUYCWKLQ

winder: We need to implement the step correctly. An incorrect step implementation like this will cause problems down the road.

algonautshant: Thanks for point to the issue. parameters restored.

PR #180, opened 2020-08-02
alternate equals

winder:
I took a stab at re-implementing the java recursive compare algorithm in go. This doesn't re-sort the array, but it does handle default / missing values, and unsorted objects nicely.


test/indexer_integration_test.go
```diff
@@ -592,33 +592,13 @@ func iUseToSearchForTransactionsWithAppIdAndToken(
...
-	if err != nil {
-		excludeBool = false
-	}
+	ic := indexerClients[indexer]
 	response, err = ic.SearchForTransactions().
 		ApplicationId(uint64(appId)).
 		Limit(uint64(limit)).
-		NotePrefix(notePrefixBytes).
-		TxType(txType).SigType(sigType).
-		TXID(txid).Round(uint64(round)).
```
winder: This is the wrong txid, it should be `txId`

test/utilities.go
```diff
@@ -71,27 +335,22 @@ func EqualJson(j1, j2 string) (ans bool, err error) {
...
-			return false, err
+			matches := differRegex.FindStringSubmatch(line)
+			if ! binaryOrStringEqual(matches[2], matches[3]) {
+				return fmt.Errorf(str)
+			}
 		}
 	}
-	// Extra layer of check. This is not expected to be true.
-	// If only difference is removed with false value property, then the difference
-	// will be qualified as SupersetMatch. If not, should be be considered equal.
```
winder, line 89: The binary/string comparison is incompatible with the superset match, so I needed to remove it.

test/utilities.go
```diff
@@ -3,30 +3,297 @@ package test
...
 	"sort"
 	"strings"
 
 	"github.com/nsf/jsondiff"
+
+	sdk_json "github.com/algorand/go-algorand-sdk/encoding/json"
+	"github.com/algorand/go-algorand-sdk/encoding/msgpack"
 )
 
+func VerifyResponse(expectedFile string, actual string) error {
```
winder, line 23: Moved the file loading in here so that I could process base64 files in one place, and call both equals functions.

winder:
Closing because this has been included in #199

PR #183, opened 2020-08-11
Update README.md

dpaks:
Optimised a Readme example.

Hello, this is my first PR here. I know this is a Readme example, but still it would be great to have optimal examples.

**Problem Area**
We were looping over wallets to find a match.
**Problem**
Even after finding a match, we still looped unnecessarily over the other wallets, if any.
**Fix**
Break the loop after a match is found.


winder:
Thanks @dpaks !

PR #187, opened 2020-10-21
Add BlockRaw method

winder:
Add BlockRaw method to fetch the raw block bytes. Minor cleanup and extensions to the common client code to support this.

### DIFF
--- a/client/v2/common/common.go
+++ b/client/v2/common/common.go
@@ -182,6 +182,7 @@ func (client *Client) Get(ctx context.Context, response interface{}, path string
 	return client.submitForm(ctx, response, path, body, "GET", false /* encodeJSON */, headers)
 }
 
+// GetRaw performs a GET request to the specific path against the server and returns the raw body bytes.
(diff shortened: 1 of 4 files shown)


client/v2/common/common.go
```diff
@@ -103,34 +103,34 @@ func mergeRawQueries(q1, q2 string) string {
 }
 
 // submitForm is a helper used for submitting (ex.) GETs and POSTs to the server
```
EvanJRichard, line 105: Can you please fix this comment to match the function name?

EvanJRichard: And does SubmitFormRaw need to be public like this? ("Yes" is a perfectly fine answer)

EvanJRichard (approved):
LGTM, noted a couple things about comments/public vs. private

client/v2/common/common.go
winder: Thanks, I don't think this needs to be public anymore. I was using it before adding "GetRaw"

PR #193, opened 2021-02-06
Adding codec struct tags to SuggestedParams struct

barnjamin:
This struct does not go over the wire so codec struct tags are not required but adding them here to make the json output match documentation


fabrice102:
That is with regards to https://developer.algorand.org/docs/build-apps/connect/#check-suggested-transaction-parameters
Currently the JSON output does not match the documentation above.

types/transaction.go
```diff
@@ -164,31 +164,30 @@ type SuggestedParams struct {
...
+	LastRoundValid Round `codec:"last-round"`
 
 	// ConsensusVersion indicates the consensus protocol version
 	// as of LastRound.
-	ConsensusVersion string
+	ConsensusVersion string `codec:"consensus-version"`
 
-	// FlatFee indicates whether the passed fee is per-byte or per-transaction
 	// If true, txn fee may fall below the MinTxnFee for the current network protocol.
 	FlatFee bool
```
jasonpaulos, line 186:
```suggestion
	FlatFee bool `codec"flat-fee"`
```

types/transaction.go
```diff
@@ -164,31 +164,30 @@ type SuggestedParams struct {
...
 
 	// ConsensusVersion indicates the consensus protocol version
 	// as of LastRound.
-	ConsensusVersion string
+	ConsensusVersion string `codec:"consensus-version"`
 
-	// FlatFee indicates whether the passed fee is per-byte or per-transaction
 	// If true, txn fee may fall below the MinTxnFee for the current network protocol.
-	FlatFee bool
+	FlatFee bool `codec"flat-fee"`
```
barnjamin, line 186: This is missing the  colon between tag and value

jasonpaulos: Thanks for catching that!

types/transaction.go
```diff
@@ -164,31 +164,30 @@ type SuggestedParams struct {
...
+	FirstRoundValid Round `codec:"first-round"`
 
 	// LastRoundValid is the final protocol round on which the txn may be committed
-	LastRoundValid Round
+	LastRoundValid Round `codec:"last-round"`
 
 	// ConsensusVersion indicates the consensus protocol version
 	// as of LastRound.
-	ConsensusVersion string
+	ConsensusVersion string `codec:"consensus-version"`
```
jasonpaulos, line 184:
```suggestion

// FlatFee indicates whether the passed fee is per-byte or per-transaction
```

types/transaction.go
```diff
@@ -164,31 +164,31 @@ type SuggestedParams struct {
...
 
-	// FlatFee indicates whether the passed fee is per-byte or per-transaction
+// FlatFee indicates whether the passed fee is per-byte or per-transaction
```
jasonpaulos, line 185:
```suggestion
	// FlatFee indicates whether the passed fee is per-byte or per-transaction
```

jasonpaulos (approved):
Thank you for this contribution! Sorry for the delay in review. This will be included in the next release.

PR #199, opened 2021-03-15
Indexer 2.3.1 support

winder:
Most of these changes were generated with a new code generator. The major differences are:
* Rather than all types in a single model file they are separate files. This is because algod types are now a subset of indexer types, so we need to overwrite some model objects written by the algod generator. This is much simpler with separate files.
* The query parameter objects have been moved from the models' package to each query's file. I tried to make these types private but the HTTP code needs them to be public to allow for marshaling. In the future, we could update the common HTTP code to allow them to be private.

Other changes
* The common HTTP code has been changed to marshal the body to a response object, if possible, even if there was an error. This is needed because we sometimes return a 404 from the indexer along with some other metadata in the body.
* Replaced the old Response / JSON-file comparison function with #180, which supports the msgpack tests.
* There were a couple of places where invalid query parameter functions have been removed because they don't exist.


client/v2/common/common.go
```diff
@@ -70,14 +70,13 @@ type InternalError error
 // extractError checks if the response signifies an error.
```
winder, line 70: This code is common to all endpoints. We now attempt to marshal the response body into the response body even if there was an error. Sometimes there is still useful information to return.

client/v2/common/models/application_responsemodels.go
```diff
@@ -2,107 +2,6 @@ package models
```
winder, line 2: Remove hand-copied types which are now generated.

client/v2/common/models/filtermodels.go
```diff
@@ -1,13 +1,5 @@
 package models
```
winder, line 1: These are the parameters to query objects. In the newly generated code they are declared inside the query file.

client/v2/common/models/responsemodels.go
```diff
@@ -1,82 +1,5 @@
 package models
```
winder, line 1: Remove more types that are now regenerated.

client/v2/common/models/shim.go
```diff
@@ -0,0 +1,10 @@
+package models
+
+// Types in this file are used to help convert between hand written code
+// that doesn't quite match the spec, and code generated from the spec.
```
winder, line 4: This was going to be too ugly to cram into the codegen. The generator maps things like `NodeStatusResponse` -> `NodeStatus` and this file connects the custom name to the generated types.

test/algodclientv2_test.go
```diff
@@ -262,7 +263,7 @@ func weMakeAPendingTransactionInformationAgainstTxidWithMax(txid string, max int
 	if err != nil {
 		return err
 	}
-	_, _, globalErrForExamination = algodClient.PendingTransactionInformation(txid).Max(uint64(max)).Do(context.Background())
+	_, _, globalErrForExamination = algodClient.PendingTransactionInformation(txid).Do(context.Background())
```
winder, line 266: Max was included before by mistake, this is a non-backwards compatible change but wasn't valid to begin with.

test/indexer_integration_test.go
```diff
@@ -552,40 +563,50 @@ func thereAreAssetsInTheResponseTheFirstIs(numAssetsInResponse, assetIdFirstAsse
...
 	ic := indexerClients[indexer]
 	var err error
-	response, err = ic.LookupApplicationByID(uint64(appid)).Do(context.Background())
-	return err
+	query := ic.LookupApplicationByID(uint64(appid))
+	if ia, err := strconv.ParseBool(includeAll); err == nil && ia {
+		query.IncludeAll(ia)
+	}
+	response, err = query.Do(context.Background())
+	// allow 404...
```
winder, line 598: Technically all of these functions could be changed to only return an error when `err` is not an HTTP error. For now, this is the only one that needs that functionality.

test/responses_unit_test.go
```diff
@@ -109,18 +107,40 @@ func weMakeAnyCallTo(client /* algod/indexer */, endpoint string) (err error) {
 				GenesisID:        sParams.GenesisID,
 				Genesishash:      sParams.GenesisHash,
 				LastRound:        uint64(sParams.FirstRoundValid),
-				MinFee:           81560,
+				MinFee:           sParams.MinFee,
```
winder, line 110: Apparently the old test input only had a single value... that is no longer true so I needed to add MinFee to SuggestedParams

test/responses_unit_test.go
```diff
@@ -109,18 +107,40 @@ func weMakeAnyCallTo(client /* algod/indexer */, endpoint string) (err error) {
...
+				TotalTransactions: total,
+			}
+		case "GetPendingTransactionsByAddress":
+			var total uint64
+			var top []types.SignedTxn
+			total, top, err = algodC.PendingTransactionsByAddress("address").Do(context.Background())
+			response = models.PendingTransactionsResponse{
+				TopTransactions:   top,
+				TotalTransactions: total,
+			}
```
winder, line 133: :point_up: Old tests supported by the new verify function.

:point_down: new endpoints

test/utilities.go
```diff
@@ -1,371 +1,310 @@
...
 	"sort"
 	"strings"
 
-	"github.com/nsf/jsondiff"
+	sdk_json "github.com/algorand/go-algorand-sdk/encoding/json"
+	"github.com/algorand/go-algorand-sdk/encoding/msgpack"
 )
 
-// EqualJson compares two json strings.
+func VerifyResponse(expectedFile string, actual string) error {
```
winder, line 18: This function has been in a PR since august, it supports a wider range of tests than the old one so I went ahead and brought it in now.

types/block.go
```diff
@@ -186,9 +186,52 @@ type (
 		// Closing amount for transaction.
```
winder, line 186: Unrelated changes... add in some missing types to the `ApplyData` type.

client/v2/indexer/lookupAccountByID.go
```diff
@@ -8,14 +8,36 @@ import (
...
+	Round uint64 `url:"round,omitempty"`
+}
+
 type LookupAccountByID struct {
-	c       *Client
-	account string
-	p       models.LookupAccountByIDParams
+	c *Client
+
+	accountId string
```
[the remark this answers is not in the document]
winder: Thanks, you're right. I'm not generating the clients yet and didn't notice this changed

test/utilities.go
```diff
@@ -1,371 +1,310 @@
...
+		//sortField := getFirstField(expected)
+		//sortArray(actualArr, sortField)
+		//sortArray(expectedArr, sortField)
+
+		// n^2 baby! Just make sure every element has a match somewhere in the other list.
+		// getting the sort function to work would be better...
+		var err error
+		for i := 0; i < expectedSize; i++ {
+			for j := 0; j < len(actualArr); j++ {
+				err = recursiveCompare(fmt.Sprintf("%s[%d]", field, i), expectedArr[i], actualArr[i])
```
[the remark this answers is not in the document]
winder: wow, yeah. nice catch.

jasonpaulos (approved):
Aside from the duplicate values array comparison issue we talked about, this looks good!

jasonpaulos:
It looks like this PR doesn't include the `/v2/transactions/{txid}` endpoint for indexer. Should that be included?

PR #201, opened 2021-03-19
Code generation for more of the http client

winder:
Use the updated go code generator to generate the entire go v2 http client.

Since we call almost every endpoint/query parameter from the integration tests I'm pretty confident that these changes maintain backwards compatibility. It's possible some of the model object types changed, we don't have any coverage for those.


test/algodclientv2_test.go
```diff
@@ -5,7 +5,6 @@ import (
 	"context"
```
winder, line 5: These changes are technically unrelated to code generation.

The first version of our unit tests took a specific response and verified a single field. This replaces the single-field checks with the JSON compare functions.

test/indexer_unit_test.go
```diff
@@ -3,33 +3,30 @@ package test
 import (
```
winder, line 3: These changes are technically unrelated to code generation.

The first version of our unit tests took a specific response and verified a single field. This replaces the single-field checks with the JSON compare functions.

test/responses_unit_test.go
```diff
@@ -51,7 +50,7 @@ func weMakeAnyCallTo(client /* algod/indexer */, endpoint string) (err error) {
 		switch endpoint {
 		case "lookupAccountByID":
 			round, something, err = indexerC.LookupAccountByID("").Do(context.Background())
-			response = models.LookupAccountByIDResponse{
+			response = models.AccountResponse{
```
winder, line 53: In order for the "decomposed" versions of the API to be compared against the actual results, the full response object must be reconstructed.

These reconstructed objects are outside the scope of the client interface, so it doesn't matter that the names have changed.

I also added some stuff that was missing from adding the full json-compare to the other 2 test file changes.

PR #204, opened 2021-04-27
Fix GetGenesis endpoint.

winder:


client/v2/common/common.go
```diff
@@ -175,6 +175,12 @@ func (client *Client) submitForm(ctx context.Context, response interface{}, path
 
 	responseErr := extractError(resp.StatusCode, bodyBytes)
 
+	// The caller wants a string
+	if strResponse, ok := response.(*string); ok {
+		*strResponse = string(bodyBytes)
+		return err
+	}
```
winder, line 182: Skip encoding and assign directly if the response type is a string.

jasonpaulos: Out of curiosity, what type is response if it's meant to be decoded as JSON?

jasonpaulos (approved):
LGTM

client/v2/common/common.go
winder: In almost every other case it's a proper object which has tags for how to serialize/deserialize. For example the asset object:

```
type Asset struct {
	// CreatedAtRound round during which this asset was created.
	CreatedAtRound uint64 `json:"created-at-round,omitempty"`

	// Deleted whether or not this asset is currently deleted.
	Deleted bool `json:"deleted,omitempty"`

	// DestroyedAtRound round during which this asset was destroyed.
	DestroyedAtRound uint64 `json:"destroyed-at-round,omitempty"`

	// Index unique asset identifier
	Index uint64 `json:"index,omitempty"`

	// Params assetParams specifies the parameters for an asset.
	// (apar) when part of an AssetConfig transaction.
	// Definition:
	// data/transactions/asset.go : AssetParams
	Params AssetParams `json:"params,omitempty"`
}
```

PR #210, opened 2021-05-21
Fix FlatFee computation.

winder:
Some asset transactions were ignoring the `FlatFee` flag. 

We have a "FlatFee" feature which determines whether the "Fee" field is multiplied by the estimated transaction size, or taken at face value. Many of the Asset builders were ignoring that flag leading us to multiply the FlatFee by the estimated size.

If FlatFee is used and Fee is set to 1000, this would cause simple asset transactions to have a fee in the .15 - .2 algo range rather than .001.

### DIFF
--- a/future/transaction.go
+++ b/future/transaction.go
@@ -11,6 +11,24 @@ import (
 // MinTxnFee is v5 consensus params, in microAlgos
 const MinTxnFee = transaction.MinTxnFee
 
+func setFee(tx types.Transaction, params types.SuggestedParams) (types.Transaction, error) {
+	if !params.FlatFee {
+		eSize, err := transaction.EstimateSize(tx)
+		if err != nil {
+			return types.Transaction{}, err
+		}
+		tx.Fee = types.MicroAlgos(eSize * uint64(params.Fee))
+	} else {
+		tx.Fee = params.Fee
+	}
+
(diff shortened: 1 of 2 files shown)


future/transaction.go
```diff
@@ -62,20 +80,7 @@ func MakePaymentTxn(from, to string, amount uint64, note []byte, closeRemainderT
...
-		}
-		tx.Fee = types.MicroAlgos(eSize * uint64(params.Fee))
-	}
-
-	if tx.Fee < MinTxnFee {
-		tx.Fee = MinTxnFee
-	}
-
-	return tx, nil
+	return setFee(tx, params)
```
algonautshant, line 83: In this case, if the flatFee is set to a higher value than the MinTxnFee, the fee will be higher than before this change. Right?

algonautshant (approved):
Looks great.
Looks like the fix goes both ways: some will be cheaper now, some more expensive. 
But Now is the correct computation.

future/transaction.go
winder: You are correct about the intention, if `FlatFee = true` then we should use `max(param.Fee, MinTxnFee)`

This was actually the behavior before as well because the fee is always initialized to params.Fee. In `setFee` I'm setting it a second time just to be sure.

PR #211, opened 2021-06-03
"Rent" larger program

shiqizng:
This PR adds a parameter, extra-pages, to ApplicationCreateTxn(). This allows user to request extra space for application programs.


shiqizng:
build is failing with msgpack decode error because it uses nightly channel version of go-algorand. It is out of sync with its master branch and it's getting fixed.  I will retry the build once that's fixed.  Locally, tests passed when built from the master branch.  

updated cucumber tests,
algorand/algorand-sdk-testing@c37a172
algorand/algorand-sdk-testing@a06d2dc

types/applications.go
```diff
@@ -132,5 +133,8 @@ func (ac *ApplicationCallTxnFields) Empty() bool {
 	if ac.ClearStateProgram != nil {
 		return false
 	}
+	if ac.ExtraProgramPages !=0 {
```
jannotti, line 136: missing space before 0

jannotti (approved):
This all seems good to me, though I'm less familiar with the go sdk than the python.

I wonder if we should have a cucumber test of sending a larger program, so we confirm that no sdk has a hard coded check against 1024.  That wouldn't go into this repo directly, of course.

shiqizng:
> This all seems good to me, though I'm less familiar with the go sdk than the python.
> 
> I wonder if we should have a cucumber test of sending a larger program, so we confirm that no sdk has a hard coded check against 1024. That wouldn't go into this repo directly, of course.

I created a program file, big_app_program.teal.tok, of ~4KB for the new test case I added.

winder:
@algojohnlee be aware that when releasing these changes it needs to be a major version update because it's not backward compatible (the public Make functions are changing).

types/applications.go
```diff
@@ -86,6 +86,7 @@ type ApplicationCallTxnFields struct {
 	GlobalStateSchema StateSchema `codec:"apgs"`
 	ApprovalProgram   []byte      `codec:"apap"`
 	ClearStateProgram []byte      `codec:"apsu"`
+	ExtraProgramPages uint32      `codec:"apep,omitempty"`
```
jdtzmn, line 89: Why `omitempty` here? I would assume all "optional" fields such as `Accounts` should be `omitempty` or none of them should.

jdtzmn (changes requested):

future/transaction.go
```diff
@@ -772,7 +783,7 @@ func MakeApplicationCallTx(
 	note []byte,
 	group types.Digest,
 	lease [32]byte,
-	rekeyTo types.Address) (tx types.Transaction, err error) {
+	rekeyTo types.Address, extraPages uint32) (tx types.Transaction, err error) {
```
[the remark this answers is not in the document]
jannotti: unlike python, that seems reasonable here.  static type checking will make people fix their code.

types/applications.go
jannotti: That sounds right. I think the "omitempty" on the _struct field covers it.

future/transaction.go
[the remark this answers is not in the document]
jdtzmn: Plus, based on Will's comment, this will be a breaking change anyway.

jannotti:
Sure, it's a "breaking change" but I would never want to introduce a change
that can silently change behaviour without an awfully good reason.

That's why I'm ok with changing the arg order in Go, where you're sure to
get a compile error because the types (and arg count) don't match, but less
willing on python, where it might even run, silently doing the wrong thing
(we'd have to reason carefully about whether the argument types would cause
a runtime error in all cases - they certainly wouldn't be a compile error!

jdtzmn:
@shiqizng Sorry for the miscommunication, but if any of the SDKs put the `extraPages` parameter last, then they all should put it last for consistency across SDKs. I [mentioned this](https://github.com/algorand/py-algorand-sdk/pull/196#discussion_r645737944) in the Python PR, but you must have missed it. Reverting the last two commits should undo the reordering.

jdtzmn (approved):

PR #216, opened 2021-06-15
Fee pooling SDK change

shiqizng:
This PR is to support creating transactions with a fee under mintxnfee since other txns in the group will be paying.

### DIFF
--- a/future/transaction_test.go
+++ b/future/transaction_test.go
@@ -742,7 +742,7 @@ func TestFee(t *testing.T) {
 			expected: 1001,
 		},
 		{
-			name:     "Flat fee does not overridden with min fee",
+			name:     "Flat fee does not get overridden with min fee",
 			flatFee:  true,
 			fee:      999,
 			expected: 999,
@@ -752,6 +752,12 @@ func TestFee(t *testing.T) {
 			flatFee:  false,
 			fee:      1,
 			expected: 1000,
+		},
+		{
+			name:     "Flat fee set to 0",
+			flatFee:  true,
+			fee:      0,
+			expected: 0,
 		},
 	}
 	addr := types.Address{}.String()
--- a/future/transaction_test.go
+++ b/future/transaction_test.go
@@ -2,13 +2,14 @@ package future
 
 import (
 	"encoding/base64"
+	"testing"
+
 	"github.com/algorand/go-algorand-sdk/crypto"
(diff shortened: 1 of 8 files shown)


future/transaction_test.go
```diff
@@ -741,10 +742,10 @@ func TestFee(t *testing.T) {
 			expected: 1001,
 		},
 		{
-			name:     "Flat fee overridden with min fee",
+			name:     "Flat fee does not overridden with min fee",
```
jdtzmn, line 745: nit: _is not_ or _does not get_

jdtzmn (changes requested):
Mostly looks good - I just left one comment. Can you also add a test case that explicitly sets the fee to 0?

jasonpaulos (changes requested):
Implementation looks good, but could you modify the descriptions for the `SuggestedParams` struct? Specifically:
* For `Fee`, it current says "Fee may fall to zero but transactions must still have a fee of at least MinTxnFee for the current network protocol." Maybe change this to "Fee may fall to zero but a group of N atomic transactions must still have a fee of at least N*MinTxnFee for the current network protocol."
* For `FlatFee`, mention that if this is true, the fee is not guaranteed to be above the network minimum.

jasonpaulos (approved):
Looks good! Thanks for making the requested changes

types/transaction.go
```diff
@@ -162,8 +162,8 @@ type TxGroup struct {
 type SuggestedParams struct {
 	// Fee is the suggested transaction fee
 	// Fee is in units of micro-Algos per byte.
-	// Fee may fall to zero but transactions must still have a fee of
-	// at least MinTxnFee for the current network protocol.
+	// Fee may fall to zero but a group of N atomic transactions must
+	// still have a fee of at least N*MinTxnFee for the current network protocol.
```
winder, line 166: Might be nice to check the minimum balance in `crypto.ComputeGroupID`

types/transaction.go
```diff
@@ -183,6 +183,7 @@ type SuggestedParams struct {
 	ConsensusVersion string
 
 	// FlatFee indicates whether the passed fee is per-byte or per-transaction
+	// If true, txn fee is may fall below the MinTxnFee for the current network protocol.
```
jdtzmn, line 186: nit: txn fee _may_ fall

jdtzmn (changes requested):

jdtzmn (approved):

PR #223, opened 2021-07-01
create v2 module

shiqizng:
v2.0.0 cannot be obtained using Go tools, 
`go get github.com/algorand/go-algorand-sdk/...@v2.0.0: github.com/algorand/go-algorand-sdk@v2.0.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2`

To fix this issue, this PR creates a v2 subdirectory including v2 go files and updated module path to github.com/algorand/go-algorand-sdk/v2/


algorandskiy:
what are `*.go-e` files? If copying files then the root must be reverted to v1.9.1 and v2.0.0 copied to v2 subfolder.
But I also wonder if we can somehow make common "internal" package(s) and share between v1 and v2 to minimize amount of duplication.

winder:
Is there a good reason not to use the suggestion here? https://therootcompany.com/blog/bump-go-package-to-v2/

shiqizng:
> Is there a good reason not to use the suggestion here? https://therootcompany.com/blog/bump-go-package-to-v2/

I followed this go blog post https://blog.golang.org/v2-go-modules.  the suggested changes seem to be identical unless I missed something.

shiqizng:
> what are `*.go-e` files? If copying files then the root must be reverted to v1.9.1 and v2.0.0 copied to v2 subfolder.
> But I also wonder if we can somehow make common "internal" package(s) and share between v1 and v2 to minimize amount of duplication. @cce thoughts?

`*.go-e` files seem to contain import path to previous version like` github.com/algorand/go-algorand-sdk/types`. I will remove these files. 

should the root be reverted to v1.9.0? according to the note here, https://github.com/algorand/go-algorand-sdk/releases, v1.9.1 and v2.0.0. are the same. @jasonpaulos

winder:
If a major version requires forking the entire codebase maybe we shouldn't do that now, since there are other things we would want to put in such a drastic change (like deprecating the legacy rest API and promoting the `future` rest API).

Can we just make a new constructor instead of adding a new parameter to the old one?

jasonpaulos:
> Can we just make a new constructor instead of adding a new parameter to the old one?

I would also be in favor of this, since this would fix the problem that necessities v2 in the first place.

shiqizng:
> If a major version requires forking the entire codebase maybe we shouldn't do that now, since there are other things we would want to put in such a drastic change (like deprecating the legacy rest API and promoting the `future` rest API).
> 
> Can we just make a new constructor instead of adding a new parameter to the old one?

will this need to be done for other SDKs, too?

jasonpaulos:
I don't believe so, since all other SDKs added extra pages as an optional parameter, but that's not possible in Go.

I think Will's proposal is to make a new set of functions to create app call txns with the extra pages parameter instead of modifying the existing ones, so we'd have new functions like `MakeApplicationCreateTxWithExtraPages`.

shiqizng:
I will open a different PR after adding these functions.

Issue #224, opened 2021-07-01
FromBase64String does not update the SignedTxn


king-iron:
Making a call to types.SignedTxn.FromBase64String (https://github.com/algorand/go-algorand-sdk/blob/develop/types/basics.go#L70) should return or update the SignedTxn. 

For example, if I first create a signed transaction with goal:

```
goal clerk send --from "[ADDR_1]" -t "[ADDR-2]" -a 100000 --fee 100 --dryrun-dump-format msgp -o tx.tx -s
cat tx.tx | base64 -w 0
```

And then attempt to use the go-algorand-sdk FromBase64String() to get it back:
```
            var st types.SignedTxn
            b64 := [TEXT FROM ABOVE]
            st.FromBase64String(b64)
```

will result in an error.  

In types.SignedTxn.FromBase64String the SignedTxn looks to be modified in place and then nothing is returned back. 

I am linking a PR which updates the returns types for FromBase64String. This is my first golang PR so I am not sure if it is best practice; kid gloves please =).

PR #225, opened 2021-07-01
Updates  FromBase64String() to correctly return the signed transaction

king-iron:
Fix for https://github.com/algorand/go-algorand-sdk/issues/224

### DIFF
--- a/types/basics.go
+++ b/types/basics.go
@@ -67,19 +67,19 @@ func ToMicroAlgos(algos float64) MicroAlgos {
 	return MicroAlgos(math.Round(algos * microAlgoConversionFactor))
 }
 
-func (signedTxn SignedTxn) FromBase64String(b64string string) (SignedTxn, error) {
+func (signedTxn *SignedTxn) FromBase64String(b64string string) error {
 	txnBytes, err := base64.StdEncoding.DecodeString(b64string)
 	if err != nil {
-		return SignedTxn{}, err
+		return err
 	}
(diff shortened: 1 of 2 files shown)


winder:
Thanks, you're right this doesn't work at all...

There's a simpler solution though, this should work if you make the receiver a pointer
```diff
-func (signedTxn SignedTxn) FromBase64String(b64string string) error {
+func (signedTxn *SignedTxn) FromBase64String(b64string string) error {
```

Looks like block needs this also...
```diff
-func (block Block) FromBase64String(b64string string) error {
+func (block *Block) FromBase64String(b64string string) error {
```

Would you be able to update these and make a small unit test?

king-iron:
Thanks @winder! Using a receiver pointer makes a lot more sense. 

I am having trouble with the unit test. Specifically, where to place it in the code base and then how to run the unit test from the makefile?

winder:
It looks like we don't have tests for this file, so you would make one named `basics_test.go`

It would look something like:
```
package types

import (
	"testing"

	"github.com/stretchr/testify/require"
)

func TestSignedTxnFromBase64String() {

}

func TestBlockFromBase64String() {

}
```

You can put a target like this in the Makefile, the existing `make unit` pulls in some integration tests that you shouldn't have to worry about for this:
```
test:    
  go test $(TEST_SOURCES_NO_CUCUMBER)
```

king-iron:
Thank you! I added a unit test in.

winder:
@king-iron thanks for the contribution!

PR #226, opened 2021-07-01
Fix - v2.0.0 cannot be used with `go mod` 

shiqizng:
v2.0.0 cannot be obtained using Go tools,
`go get github.com/algorand/go-algorand-sdk/...@v2.0.0: github.com/algorand/go-algorand-sdk@v2.0.0: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2`

 adding extraPages as a new parameter to MakeApplicationCallTx(...) was a breaking change which required upgrading the package to v2.0.0.   This PR fixes this issue.

### DIFF
--- a/future/transaction.go
+++ b/future/transaction.go
@@ -534,7 +534,7 @@ func MakeApplicationCreateTx(
 	note []byte,
 	group types.Digest,
 	lease [32]byte,
-	rekeyTo types.Address, extraPages ...uint32) (tx types.Transaction, err error) {
+	rekeyTo types.Address) (tx types.Transaction, err error) {
 
 	oncomp := types.NoOpOC
 	if optIn {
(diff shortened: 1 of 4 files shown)


algorandskiy (approved):
Looks good. @algojohnlee Please remove v2.0.0 release, we found a way to update API without breaking backward compatibility, so no need to bump a major version now.

winder (changes requested):
This is clever, but too much of a hack for me... I don't think we should put something like this in the SDK

tsachiherman:
> This is clever, but too much of a hack for me... I don't think we should put something like this in the SDK

I agree with Will on this one. And not because I think it's a hack.

This would make the old code compile just fine for, but it's still an interface change. If the user would pass more than a single extra parameter to `MakeApplicationCreateTx`, these parameters would be ignored. If we will want to add yet-another-parameter to the method, it would break as well.

I don't have a clear vision to what the ideal solution in this case is; few thoughts - 
1. Automatically detect the program size from the arguments and adjust the number of extra pages.
2. Extend the interface by creating another `MakeApplicationCreateTx` variation. i.e. MakeApplicationCreateTxEx, MakeApplicationCreateTx2, etc.
3. Create a parametrized version of MakeApplicationCreateTx

algorandskiy:
> This is clever, but too much of a hack for me... I don't think we should put something like this in the SDK

Well, how about Python/JS SDKs? We've been adding additional params all the time not saying "this is a hack" or "it's still an interface change".
I think this is appropriate solution for not bumping major version, only need to enforce there is only zero or one variadic param is allowed.

shiqizng:
created new functions to accept extraPages parameter that doesn't break the previous versions.

winder (approved):
@pzbitskiy maybe you're right and the SDKs aren't being handled equally. For python, I think it's a little different since we have named parameters. Not sure about javascript.

I share your frustration. When this future package was being added I pitched several solutions that would support optional/new parameters, but they weren't implemented:
https://algorand.atlassian.net/wiki/spaces/FDE/pages/491716666/SuggestedParams+Thoughts

For now, I think the updated solution with additional make functions is the best we can do without making this a larger project.

shiqizng:
Will using a builder pattern for txn creations be better?

winder:
I happen to like the builder pattern since it lets you do things like have multiple setters for the same field (i.e. address as a string or a type.Address object).

But for go, and with the current team priorities, a parameter struct gives us all of the maintainability benefits for a fraction of the work.

future/transaction.go
```diff
@@ -534,7 +534,7 @@ func MakeApplicationCreateTx(
 	note []byte,
 	group types.Digest,
 	lease [32]byte,
-	rekeyTo types.Address, extraPages uint32) (tx types.Transaction, err error) {
+	rekeyTo types.Address) (tx types.Transaction, err error) {
```
jasonpaulos, line 538: Nit: to reduce (a small amount of) code duplication, could you make the body of this function just call `MakeApplicationCreateTxWithExtraPages` with 0 extra pages?

jasonpaulos (approved):
I left 1 minor comment, but this looks good to merge with or without that

PR #228, opened 2021-07-08
Use correct go version in CI

jasonpaulos:
I noticed that the Dockerfile for this repo always uses Go 1.13, which means our Travis build matrix for multiple Go versions is not working.

So I made some modifications to pull from a base image with the same version of Go that the system currently uses. I also added Go versions 1.15 and 1.16 to our build matrix.


winder (approved):
Nice

PR #230, opened 2021-07-12
Signing support for rekeying to LogicSig/MultiSig account

jasonpaulos:
Currently this SDK has strict checks in `SignMultisigTransaction` and `SignLogicsigTransaction` that prevent transactions with an unexpected sender address from being signed. Since it's possible to rekey other accounts to LogicSig and MultiSig accounts, these types of accounts should be able to sign transactions from any address. This PR adds that functionality.

MultiSig changes:
* `SignMultisigTransaction` can now properly sign transactions from any address.

LogicSig changes:
* `SignLogicsigTransaction` can now properly sign transactions from any address, **except if the LogicSig is delegated to a non-multisig account and the sender is not the delegating account**. This case is impossible to support because the current `LogicSig` struct does not keep track of the public key which provided the signature, so we cannot add the correct `AuthAddr` to the signed transaction.
* Introduce the struct `LogicSigAccount`, which contains a `LogicSig` and a the public key that signed it, if any.
* Introduce the function `SignLogicSigAccountTransaction`, which signs transactions from a given `LogicSigAccount`. Unlike `SignLogicsigTransaction`, this function can sign transactions from any address without limitations.
* Deprecate `MakeLogicSig` in favor of using the `LogicSigAccount` struct.

Other changes:
* Add `AccountFromPrivateKey`, a missing method for obtaining an Account from only a private key.


types/signature.go
```diff
@@ -50,13 +50,29 @@ type LogicSig struct {
 	// OR hashed to be the Address of an account.
 	Logic []byte `codec:"l"`
 
-	Sig  Signature   `codec:"sig"`
+	// The signature of the account that has delegated to this LogicSig, if any
+	Sig Signature `codec:"sig"`
+	// The key that provided Sig, if any
+	SigKey ed25519.PublicKey `codec:"sigkey"`
```
algorandskiy, line 56: imo adding a new field into a struct that is going to be serialized, and removing it before serializing is error-prone. Please consider alternative approach - wrapping this canonical struct or passing it along with `LogicSig` type

crypto/crypto.go
```diff
@@ -445,39 +450,74 @@ func VerifyLogicSig(lsig types.LogicSig, sender types.Address) (result bool) {
 		return false
 	}
 
-	result = false
 	toBeSigned := programToSign(lsig.Logic)
-	// logic sig, compare hashes
-	if !hasSig && !hasMsig {
-		result = types.Digest(sha512.Sum512_256(toBeSigned)) == types.Digest(sender)
-		return
-	}
```
algorandskiy, line 454: why this did go? Shouldn't non-delegated logicsig continue to be prechecked?

crypto/crypto.go
```diff
@@ -445,39 +450,74 @@ func VerifyLogicSig(lsig types.LogicSig, sender types.Address) (result bool) {
...
 
 	if hasSig {
-		result = ed25519.Verify(sender[:], toBeSigned, lsig.Sig[:])
-		return
+		return ed25519.Verify(lsig.SigKey[:], toBeSigned, lsig.Sig[:])
 	}
 
-	result = VerifyMultisig(sender, toBeSigned, lsig.Msig)
```
algorandskiy, line 461: does this old code assumes `sender == msigAccount`` ?

crypto/crypto.go
jasonpaulos: The only purpose of this check is to verify that the sender is the LogicSig's hashed program address. This check is no longer needed, since we want to support signing txns for accounts rekeyed to LogicSigs.

crypto/crypto.go
jasonpaulos: Yes it does. I believe this was written before rekeying.

jdtzmn (approved):
Looks good, but I don't have any experience with the go SDK so it's probably worth having another set of eyes look it over. However, I did find an existing typo that should be changed.

crypto/account.go
```diff
@@ -50,6 +51,35 @@ func GenerateAccount() (kp Account) {
 	return
 }
 
+// AccountFromPrivateKey derives the remaining Account fields from only a private key
+func AccountFromPrivateKey(sk ed25519.PrivateKey) (account Account, err error) {
+	if len(sk) == ed25519.SeedSize {
```
winder, line 56: How does this happen? It looks like `NewKeyFromSeed` accepts a byte array, so if we've already converted the seed into `ed25519.PrivateKey` doesn't that mean it's already gone through this conversion?

crypto/account.go
```diff
@@ -145,3 +180,189 @@ func LogicSigAddress(lsig types.LogicSig) types.Address {
...
+		addr, err = msigAccount.Address()
+		return
+	}
+
+	addr = LogicSigAddress(lsa.Lsig)
+	return
+}
+
+// SignTransaction signs a transaction using this LogicSigAccount.
+func (lsa LogicSigAccount) SignTransaction(tx types.Transaction) (txid string, stxBytes []byte, err error) {
```
winder, line 360: This matches the account object with the sign function used by Java, but I thought the other SDKs use standalone functions.

Maybe this should be `crypto.LogicSigSignTransaction(lsig, tx)`?

Or maybe SignTransaction could be changed to accept the go `crypto.PrivateKey`, and `LogicSig` can implement that interface?

That said, I'm not really familiar with using smart contracts, so maybe this doesn't work.

winder: It looks like you have `SignLogicsigTransaction` also, so there are two ways to sign?

winder: Looking a little further, there is a `Signer` interface. Maybe there is a way to use that?

I understand that we're implementing Knapsack soon, so figuring out this interface could be really helpful.

winder (commented):
I left some comments. The code looks more or less fine, but I'll defer to you guys for the functionality.

crypto/account.go
jasonpaulos: Good point, I misinterpreted some things about `ed25519.PrivateKey` when I wrote this. I will update this function to just return an error if the length of `sk` is not `ed25519.PrivateKeySize`.

crypto/account.go
jasonpaulos: My intent was to allow the existing `SignLogicsigTransaction` to still work for compatibility, but to provide an alternative way to sign using the new `LogicSigAccount` object instead of a `LogicSig`. Originally I didn't want to introduce a new `SignLogicSigAccountTransaction` because it sounds too similar, but now I think this might be the best compromise.

jasonpaulos: I'd be interested in having our account objects implement a common `Signer` interface (thought maybe we'd want a more specific `TransactionSigner` interface), but that seems like a bigger change that should happen in a different PR. 

Plus, I'm not aware of the details of how knapsack signatures will be used to anticipate how the SDK can support them yet.

ryanRfox:
Yes, this looks good to me. I look forward to a future `TransactionSigner` interface as well.

PR #232, opened 2021-07-16
Implemented WaitForConfirmation function

DamianB-BitFlipper:
When using the Golang SDK, I oftentimes find myself copying the same code to WaitForConfirmation() of a pending transaction.

The WaitForConfirmation() waits for a pending transaction to be accepted and confirmed by the network. Until then, it blocks. I have implemented this function with a timeout feature as well that will error if the timeout is achieved.

I am unsure what to do about the CHANGELOG


client/v2/algod/waitForConfirmation.go
```diff
@@ -0,0 +1,61 @@
...
+
+// WaitForConfirmation waits for a pending transaction to be accepted by the network
+type WaitForConfirmation struct {
+	c *Client
+
+	// The ID of the pending transaction to wait for
+	txid string
+
+	// The number of rounds to block before existing with an error. If zero, there is no timeout
+	timeout uint64
```
algorandskiy, line 19: please rename to reflect rounds and not time

winder:
This is generated code, so the next time we re-generate the REST client the changes would be reverted.

I think we need to add this functionality to the algod REST client rather than implement it in each SDK.

DamianB-BitFlipper:
What do you mean by "generated code'?

winder:
We have a tool that generates these source files based on an OpenAPI specification of the REST endpoints, currently the templates are here:
https://github.com/algorand/generator/tree/master/go_templates

DamianB-BitFlipper:
I understand. You mean that the JS SDK is generated. Is this for JS or all SDKs?

I first interpreted that we should add this to the algod REST API which I found peculiar. But this is not the case.

jasonpaulos:
> I understand. You mean that the Go SDK is generated. Is this for Go or all SDKs?

Most of our SDKs use some amount of code generation for the algod and indexer clients. For instance, everything in the `client/v2/algod` folder is generated code. If you add this new function there, it will be overwritten the next time we generate the algod client.

Instead, this is not a perfect solution, but I would suggest making a new file in the `future` folder which implements the wait for confirmation function, similar to your Python PR: https://github.com/algorand/py-algorand-sdk/pull/214

future/waitForConfirmation.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// The ID of the pending transaction to wait for
+	txid string
+
+	// The number of rounds to block before existing with an error. If zero, there is no timeout
+	waitRounds uint64
+}
+
+// Do performs the HTTP request
+func (s *WaitForConfirmation) Do(ctx context.Context, headers ...*common.Header) (response models.NodeStatus, err error) {
```
jasonpaulos, line 24:
```suggestion
func WaitForConfirmation(c *algod.Client, txid string, waitRounds uint64, ctx context.Context, headers ...*common.Header) (response models.PendingTransactionInfoResponse, err error) {
```

Can you change the signature to this? Since this is no longer part of the algod package, it can just be a standalone function.

jasonpaulos (changes requested):

jasonpaulos (approved):
Looks great, thank you!

PR #238, opened 2021-08-31
remove contract template feature

brianolson:
part of https://github.com/algorand/go-algorand-internal/issues/1415


michaeldiamant:
Closing - We intend to pick up work here in the near future, but we can do so with a fresh PR.  Thanks to @brianolson for providing a starting point.

PR #243, opened 2021-09-21
Update Travis Environment Variables

algobarb:
## Summary
Due to a Travis security issue recently, we'll be rotating and re-encrypting Travis variables.

## Test
Making sure Travis passes with the new variables.


algojack:
are the encrypted values really necessary? or is there a better way like env vars or pulling from a secret store? (not sure how we handle secrets in a public repo like this without allowing anyone to use them, but i think there might be a better way).

algobarb:
I moved the environment values from travis yml file and into the Travis Project settings instead.

algojack (approved):
lgtm

algobarb:
For some reason travis mixed up this PR with the forked pr test I had.. so I'll recreate this PR.

PR #247, opened 2021-09-27
Add ABI-encoding feature

ahangsu:
This PR adds the corresponding feature to #236.

We added a wrapper around the implementation in go-algorand merged in https://github.com/algorand/go-algorand/pull/2807, and exported necessary interface in the sdk.


encoding/abi/abi.go
```diff
@@ -0,0 +1,223 @@
+package abi
+
+import (
+	"math/big"
+
+	"github.com/algorand/go-algorand/data/abi"
+)
+
+type Value struct {
```
jasonpaulos, line 9: @algorandskiy mentioned using struct embedding (e.g. `type Value struct { abi.Value }`) to avoid having to declare methods again, such as `Encode()`.

But would a simple type assignment work as well? E.g. `type Value = abi.Value`

algorandskiy: Wow I looks like it works @ahangsu 
I tried type aliasing (`type Value abi.Value`) and it did not work

ahangsu: That works, it can also remove the redundant method wrapper, leaving only necessary constructor, `Decode` and `TypeFromString`. Nice one :-)

algorandskiy (approved):

PR #248, opened 2021-09-27
Support AVM 1.0

### LINKED ISSUE #246: Support AVM 1.0
## Summary
Update this SDK to work with features from AVM 1.0.

## Scope
Specifically:
* Update TEAL langspec & add corresponding tests
* Make sure algod REST endpoints can support returning logs and inner transactions (regenerate REST client code).
* Add a way to get app address from app ID (and test it).

## Urgency
AVM 1.0 just got enabled in BetaNet, so good to do this as soon as possible.

jasonpaulos:
- [x] Update TEAL langspec to v5
- [x] Add function `GetApplicationAddress`
- [x] Regenerate REST API types
- [x] Support new logs cucumber test
- [x] Enable new cucumber test

Closes #246.


logic/logic_test.go
```diff
@@ -160,3 +161,24 @@ func TestCheckProgramV4(t *testing.T) {
 	err = CheckProgram(program, args)
 	require.NoError(t, err)
 }
+
+func TestCheckProgramV5(t *testing.T) {
```
shiqizng, line 165: I think there are a few other opcodes in V5 like cover, uncover, log, etc. May be good to add tests for those, too.

crypto/crypto_test.go
```diff
@@ -730,3 +730,11 @@ func TestTealSign(t *testing.T) {
 	verified2 := TealVerify(pk, data, addr, sig1)
 	require.False(t, verified2)
 }
+
+func TestGetApplicationAddress(t *testing.T) {
+	appID := uint64(77)
+	expected := "PCYUFPA2ZTOYWTP43MX2MOX2OWAIAXUDNC2WFCXAGMRUZ3DYD6BWFDL5YM"
```
shiqizng, line 736: how is this expected value derived?

client/v2/common/models/application_logs_response.go
```diff
@@ -0,0 +1,17 @@
...
+
+	// CurrentRound round at which the results were computed.
+	CurrentRound uint64 `json:"current-round"`
+
+	// LogData
+	LogData []ApplicationLogData `json:"log-data,omitempty"`
+
+	// NextToken used for pagination, when making another request provide this token
+	// with the next parameter.
+	NextToken string `json:"next-token,omitempty"`
```
shiqizng, line 16: how is the nextToken calculated?

crypto/crypto_test.go
jasonpaulos: It's from https://github.com/algorand/go-algorand/blob/412aef5ed061d96d02edfc6fa9ead6773907e66c/data/basics/userBalance_test.go#L254

client/v2/common/models/application_logs_response.go
jasonpaulos: It's supposed to treated as an opaque string that's only used to pass into the `LookupApplicationLogsByID.Next` method.

In practice I believe it's the TxID of the next transaction response, calculated by postgres: https://github.com/algorand/indexer/blob/9040a2e6c7b79ebda11f1531afd72020963c7849/api/handlers.go#L745

logic/logic_test.go
jasonpaulos: Good idea

shiqizng (approved):
looks good to me

PR #256, opened 2021-10-19
New keyreg txn field

shiqizng:
This PR adds a new field, stateProofKey, to key reg txn to match the changes in algod.


client/algod/models/models.go
```diff
@@ -462,6 +467,11 @@ type KeyregTransactionType struct {
 	//
 	// required: false
 	VoteKeyDilution uint64 `json:"votekd"`
+
+	// BlockProofPK is the block proof public key used in key registration transactions
+	//
+	// required: false
+	BlockProofPK types.Verifier `json:"bprfkey"`
```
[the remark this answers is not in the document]
algoidan: the word PK might be misleading, because this is a Merkle root commitment.
maybe we should the phrase ID instead?

client/v2/common/models/error_response.go
```diff
@@ -3,7 +3,7 @@ package models
 // ErrorResponse an error response with optional data field.
 type ErrorResponse struct {
 	// Data
-	Data *map[string]interface{} `json:"data,omitempty"`
+	Data string `json:"data,omitempty"`
```
jasonpaulos, line 6: I'm confused why this changed...

future/transaction.go
```diff
@@ -139,6 +139,73 @@ func MakeKeyRegTxn(account string, note []byte, params types.SuggestedParams, vo
...
+// - params is typically received from algod, it defines common-to-all-txns arguments like fee and validity period
+// KeyReg parameters:
+// - votePK is a base64-encoded string corresponding to the root participation public key
+// - selectionKey is a base64-encoded string corresponding to the vrf public key
+// - voteFirst is the first round this participation key is valid
+// - voteLast is the last round this participation key is valid
+// - voteKeyDilution is the dilution for the 2-level participation key
+// - nonpart is an indicator marking a key registration participating or nonparticipating
+// - stateProofPK is a base64-encoded string corresponding to the block proof public key
+func MakeKeyRegTxnV2(account string, note []byte, params types.SuggestedParams, voteKey, selectionKey string, voteFirst, voteLast, voteKeyDilution uint64, nonpart bool, stateProofPK types.Verifier) (types.Transaction, error) {
```
jasonpaulos, line 154: I think `MakeKeyRegTxnWithStateProofKey` would be a better name.

Also, similar to `voteKey` and `selectionKey`, can `stateProofPK` be a string instead, then you convert it to the right type in the function? Also you might want to put it right after `selectionKey`, since this is an entirely new function and the order of arguments is more flexible.

test/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG GO_IMAGE=golang:1.13-stretch
+ARG GO_IMAGE=golang:1.14
```
jasonpaulos, line 1: Why change this?

types/basics.go
```diff
@@ -55,6 +59,9 @@ type MasterDerivationKey [masterDerivationKeyLenBytes]byte
 // Digest is a SHA512_256 hash
 type Digest [hashLenBytes]byte
 
+// Verifier is a state proof
+type Verifier [KeyStoreRootSize]byte
```
jasonpaulos, line 63: Same comment about the name as https://github.com/algorand/java-algorand-sdk/pull/266#discussion_r740651941

client/v2/common/models/error_response.go
shiqizng: this is generated by the code generator. This is the struct in algod, https://github.com/algorand/go-algorand/blob/master/daemon/algod/api/server/v2/generated/types.go#L317

test/docker/Dockerfile
shiqizng: I got an error saying that 1.14 is the minimum required golang version to build algod. I believe this requirement is enforced here, https://github.com/algorand/go-algorand/blob/master/scripts/get_golang_version.sh

jasonpaulos: Ok, that makes sense

future/transaction.go
```diff
@@ -139,6 +139,81 @@ func MakeKeyRegTxn(account string, note []byte, params types.SuggestedParams, vo
 	return setFee(tx, params)
 }
 
+// MakeKeyRegTxnV2 constructs a keyreg transaction using the passed parameters.
```
jasonpaulos, line 142: nit: still says `MakeKeyRegTxnV2`

PR #258, opened 2021-11-01
ABI Interaction

algoidurovic:
This PR includes the implementation of Method, Contract, and Interface defined in the ABI specification as well as an implementation of AtomicTransactionComposer for constructing tx groups composed of regular transactions and method call transactions.

There are a few deviations from the spec:
-The spec requires "a way to calculate the number of transactions needed to invoke a method from the in-memory object". My understanding is that this is always exactly 1. If transactions are used as arguments to AddMethodCall() then those won't be accessible by the Method object itself. If this is required then it should be provided by the transaction composer.
-**TransactionSigner args:** the pattern of passing an array of transactions and indexesToSign doesn't seem very consistent with the TransactionWithSigner objects because they contain only one transaction for a given signer. This inconsistency leads to a somewhat clunky implementation of GatherSignatures(). It's not a big deal as far as functionality goes though.
-**Use of promises in AtomicTransactionComposer:** the spec defines certain invariants like "the composer's status will be at least SIGNED after executing this method", which can lead to an incorrect state if the method returns a promise which fails to resolve quickly or runs into an error. I elected to remove the use of promises to avoid this issue and simplify the implementation (avoiding shared state access).
-**OnComplete param to AddMethodCall():** I removed this argument because I assumed the implementation should only handle regular method calls. If this isn't the case then I should be able to add it back without too much difficulty.

**EDIT: The above points have been addressed**

Testing: unit tests have been created to tests all functionality, except for methods Submit() and Execute(), which will be testing in the future with cucumber.


future/interactions.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	// signature must include method name
+	if openIdx == 0 {
+		return Method{}, errors.New("Method signature has invalid format")
+	}
+
+	name := methodStr[:openIdx]
+	stringArgs := strings.Split(methodStr[openIdx+1:closeIdx], ",")
+	returnType := methodStr[closeIdx+1:]
```
ahangsu, line 54: It seems to be that it does not allow tuple value existing in method signature?

future/interactions.go
```diff
@@ -0,0 +1,114 @@
...
+		if err != nil {
+			return Method{}, err
+		}
+	}
+
+	_, err := abi.TypeOf(returnType)
+	if err != nil {
+		return Method{}, err
+	}
+
```
ahangsu, line 67: nit: consider that `void` is also allowed in ABI method return.

future/interactions.go
algoidurovic: Good catch, fixed

future/interactions.go
```diff
@@ -0,0 +1,166 @@
...
+	// signature must include method name
+	if openIdx == 0 {
+		return Method{}, errors.New("Method signature has invalid format")
+	}
+
+	name := methodStr[:openIdx]
+	namePattern := "[_a-zA-Z][_a-zA-Z0-9]+"
+	_, err := regexp.MatchString(namePattern, name)
```
ahangsu, line 69: nit: Pavel taught me before that you can put this in `init` and precompile the regex, which can save some constant in perf (also we might do not want `_` in leading symbol)

future/atomicTransactionComposer_test.go
```diff
@@ -0,0 +1,331 @@
...
+	"crypto/ed25519"
+	"testing"
+
+	"github.com/algorand/go-algorand-sdk/crypto"
+	"github.com/algorand/go-algorand-sdk/mnemonic"
+	"github.com/algorand/go-algorand-sdk/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMakeBasicAccountTransactionSigner(t *testing.T) {
```
algochoi, line 13: Nice tests

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,500 @@
...
+	for _, abiArg := range abiArgs {
+		encodedArg, err := abiArg.AbiType.Encode(abiArg.RawValue)
+		if err != nil {
+			return err
+		}
+
+		encodedAbiArgs = append(encodedAbiArgs, encodedArg)
+	}
+
+	tx, err := MakeApplicationNoOpTx(
```
algochoi, line 295: I initially and naively thought this was okay (my bad!), but now I do think we should follow the specs and use a `MakeApplicationCallTx` and accept the `OnComplete` params. Open to counterarguments.

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,500 @@
...
+		default:
+			return errors.New("MethodArg must be either ABIValue or TransactionSigner")
+		}
+	}
+
+	if atc.Count() == MAX_GROUP_SIZE {
+		return fmt.Errorf("REACHED MAX_GROUP_SIZE: %d", MAX_GROUP_SIZE)
+	}
+
+	for _, abiArg := range abiArgs {
```
algochoi, line 286: I think we also need to consider a case here where there are greater or equal to 15 application arguments -- in that case:

```If a method has more than 14 arguments, the remaining arguments will be encoded as a tuple in the final application call argument, accessible from TEAL as txn ApplicationArgs 15.```

**Review by algochoi — COMMENTED** [review]
This is looking good, just to address some of your higher level comments:

- I found out this recently, but the method signature can also have transactions inside them, e.g. `txcalls(pay, bool, pay, axfer, byte)void`. The supported transaction "keywords" can be found [here](https://github.com/algorand/go-algorand-sdk/blob/43196b16ffcb3d5a4ef1f22f2b0445c8e7bb9b6a/types/basics.go#L14). This also lends itself for calculating the number of transactions that are invoked by this method call, e.g. `pay` + `pay` + `axfer` + `method_itself` = 4 for the `txcalls` example method. 
- Blocking calls make sense for `submit()` and `execute()`.

File: future/atomicTransactionComposer.go [hunk 7]
```diff
@@ -0,0 +1,540 @@
...
+* @returns A promise that, upon success, resolves to an object containing the confirmed round for
+*   this transaction, the txIDs of the submitted transactions, and an array of results containing
+*   one element for each method call transaction in this group. If a method has no return value
+*   (void), then the method results array will contain null for that method's return value.
+ */
+func (atc *AtomicTransactionComposer) Execute(client *algod.Client) (int, []string, []ABIValue, error) {
+	if atc.status > SUBMITTED {
+		return 0, nil, nil, errors.New("status is already committed")
+	}
+
```
ahangsu, line 484: nit: if I am not understanding it wrongly, this should update status to `committed`, right?

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,540 @@
...
+/** Represents an unsigned transactions and a signer that can authorize that transaction. */
+type TransactionWithSigner struct {
+	/** An unsigned transaction */
+	Txn types.Transaction
+	/** A transaction signer that can authorize txn */
+	Signer TransactionSigner
+}
+
+type MethodArgument interface {
```
jasonpaulos, line 118: What's the purpose of this interface and the `ABIValue` type? It seems like it just adds more friction to calling a method, since if I want to call `add(uint32,uint32)uint64` I have to make the args array like this:

```go
// get the ABI type, which is redundant since it's already in the Method object
uint32Type, err := abi.TypeOf("uint32")

args := []MethodArgument{
	ABIValue{
		AbiType:  uint32Type,
		RawValue: uint32(1),
	},
	ABIValue{
		AbiType:  uint32Type,
		RawValue: uint32(2),
	},
}
```

Ideally the caller could just do `args := []interface{}{ uint32(1), uint32(2) }`. I guess you need to get the `abi.Type` from somewhere, so maybe a method on `Arg` could be added to parse and return this?

jasonpaulos (changes requested):
Great work, the code is easy to understand and I appreciate the tests. I did find some issues though, and I realized there are some problems with the current spec.

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,540 @@
...
+	/** The current status of the composer. The status increases monotonically. */
+	status AtomicTransactionComposerStatus
+
+	/** The transactions in the group with their respective signers. If status is greater then BUILDING
+	then this slice cannot change. */
+	transactions []TransactionWithSigner
+
+	/** Method calls constructed from information passed into AddMethodCall(). Used in Execute() to extract
+	return values. */
+	methodCalls []Method
```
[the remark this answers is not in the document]
algochoi: +1 Good idea
The `Method`s lining up with `transaction` is nice and we don't have to separately keep track of the `method` txIDs, since we already need to query the logs for each method call in `execute()`.

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,540 @@
...
+	}
+
+	txinfo, err := WaitForConfirmation(client, atc.txids[0], 0, context.Background())
+	if err != nil {
+		return 0, nil, nil, err
+	}
+
+	var returnValues []ABIValue
+	for i, txid := range atc.txids {
+		if atc.transactions[i].Txn.Type != types.ApplicationCallTx {
```
algochoi, line 497: I *think* we should be safe and refer back to our `Method` map and check if the `i` is in the map. I believe `ApplicationCallTx` could be sent as a regular transaction (as part of a `TransactionWithSigner`) because we allow `appl` calls as method arguments.

future/atomicTransactionComposer.go
jasonpaulos: Ah I see that you may have copied the `MethodArgument` and `ABIValue` types from the spec doc. For this SDK they likely don't make sense for the reasons I mentioned about.

future/atomicTransactionComposer.go
jasonpaulos: Spec has been updated, see `ABIResult `: https://gist.github.com/jasonpaulos/a810abe7e86d43840d14445718565a9a

future/interactions.go
```diff
@@ -0,0 +1,179 @@
...
+
+func MethodFromSignature(methodStr string) (Method, error) {
+	openCnt := strings.Count(methodStr, "(")
+	if openCnt == 0 {
+		return Method{}, errors.New("method signature is missing an open parenthesis")
+	}
+
+	openIdx := strings.Index(methodStr, "(")
+	name := methodStr[:openIdx]
+	match, err := regexp.MatchString(`[_a-zA-Z][_a-zA-Z0-9]+`, name)
```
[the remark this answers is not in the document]
ahangsu: One minor stuff: `_` is not allowed in the first symbol.

abi/abi.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	strTuple := "(" + argumentTypes[0].String()
+	for i := 1; i < len(argumentTypes); i++ {
+		strTuple += "," + argumentTypes[i].String()
+	}
+	strTuple += ")"
+
+	return TypeOf(strTuple)
+}
```
ahangsu, line 37: I wonder if this https://github.com/algorand/go-algorand/pull/3088/commits/fac68f3cc449672acb9d963c8c19fcfea143250c might help, if it got merged, then you can use the exported function from `go-algorand`'s abi.

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,540 @@
...
+* Note: a group can only be submitted again if it fails.
+*
+* @returns A promise that, upon success, resolves to a list of TxIDs of the submitted transactions.
+ */
+func (atc *AtomicTransactionComposer) Submit(client *algod.Client) ([]string, error) {
+	if atc.status > SUBMITTED {
+		return nil, errors.New("status must be SUBMITTED or lower in order to call Submit()")
+	}
+
+	if atc.status == SUBMITTED {
```
jasonpaulos, line 443: This should not be here, so that resubmission is allowed if the transactions were not committed.

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,681 @@
...
+	/**
+	 * The raw bytes of the return value from the ABI method call. This will be empty if the method
+	 * does not return a value (return type "void").
+	 */
+	RawReturnValue []byte
+	/**
+	 * The return value from the ABI method call. This will be undefined if the method does not return
+	 * a value (return type "void"), or if the SDK was unable to decode the returned value.
+	 */
+	ReturnValue abi.Value
```
jasonpaulos, line 190: This could be just `interface{}`, I don't see the advantage to using `abi.Value`

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,681 @@
...
+/**
+ * This type represents a function which can sign transactions from an atomic transaction group.
+ * @param txnGroup - The atomic group containing transactions to be signed
+ * @param indexesToSign - An array of indexes in the atomic transaction group that should be signed
+ * @returns A promise which resolves an array of encoded signed transactions. The length of the
+ *   array will be the same as the length of indexesToSign, and each index i in the array
+ *   corresponds to the signed transaction from txnGroup[indexesToSign[i]]
+ */
+type TransactionSigner interface {
+	SignTransactions(txGroup []types.Transaction, indexesToSign []int) ([][]byte, []string, error)
```
jasonpaulos, line 24: I'm leaning toward suggesting that this method should not return TxIDs as well. The reason being TxIDs can be derrived from the signed transactions, and if they're also returned we open the door to having inconsistent data (e.g. if a buggy signer returns a TxID that does not match the signed transaction).

What is this method returned `[]types.SignedTxn` instead of `[][]byte` and `[]string`? Then it would be easy to calculate the TxID for each signed transaction.

jasonpaulos (commented):
Here are my comments from the implementation, I haven't looked at the test implementation yet.

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,681 @@
...
+*
+* An error will be thrown if the composer's status is not BUILDING, if adding this transaction
+* causes the current group to exceed MAX_GROUP_SIZE, or if the provided arguments are invalid
+* for the given method.
+ */
+func (atc *AtomicTransactionComposer) AddMethodCall(
+	params AddMethoCallParams,
+	/** A transaction signer that can authorize this application call from sender */
+	signer TransactionSigner) error {
+
```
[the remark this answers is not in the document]
algoidurovic: It seemed like a logical separation to me since all the other params are part of the appCall tx while the signer exists outside of the tx. I don't feel too strongly so I don't mind changing it.

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,681 @@
+package future
```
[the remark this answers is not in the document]
algoidurovic: The tx composer implements relies on features provided by the future package so it should also be in future until that changes. Otherwise, we'd have code in the transactions package relying on future as a dependency.

abi/abi.go
```diff
@@ -0,0 +1,33 @@
+package abi
+
```
ahangsu, line 2: nit: I think it would be better to put `abi` in `encoding` dir, but if there's any specific reason to move it out, let me know.

(BTW, did you up the version of `go-algorand`? It seems you are still writing `MakeTupleType` again.)

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,578 @@
...
+
+	/** The transaction contexts in the group with their respective signers. If status is greater then
+	 *  BUILDING then this slice cannot change.
+	 */
+	txContexts []transactionContext
+
+	// caches that will be populated when respective contents are calculated
+	cachedTxWithSigners []TransactionWithSigner
+	cachedStxs          [][]byte
+	cachedTxIDs         []string
```
jasonpaulos, line 121: I have some thoughts on these cache objects:
* `cachedTxWithSigners`: this can be calculated directly from `txContexts`, so I don't think caching this is worth the extra complexity
* `cachedStxs`: the signed transactions are already cached in the `stxBytes` field in `transactionContext`, so I don't see the point in also having this array
* `cachedTxIDs`: this is a benefitial cache since it avoids calculating more hashes than necessary, but I think it would be better implemented as a field in `transactionContext`. Then you could make `transactionContext.txID()` returned the cached version if it exists, otherwise calculate it. As currently implemented, the call on line 526 bypasses the cache, but it would be properly handled by this suggestion.

test/applications_integration_test.go
```diff
@@ -492,6 +498,77 @@ func theConfirmedPendingTransactionByIDShouldHaveAStateChangeForToIndexerShouldA
...
+			return actualResult.DecodeError
+		}
+
+		expectedBytes, err := base64.StdEncoding.DecodeString(expectedResult)
+		if err != nil {
+			return err
+		}
+
+		if abiMethod.Returns.AbiType == "void" {
+			continue
```
jasonpaulos, line 547: Perhaps make sure that `expectedResult` is empty in this case?

test/steps_test.go
```diff
@@ -2058,3 +2102,457 @@ func tealCheckDryrun(result string) error {
...
+	}
+	return true
+}
+
+func checkEqualInterfaces(interface1, interface2 future.Interface) bool {
+	if interface1.Name != interface2.Name {
+		return false
+	}
+
+	for i, method := range interface1.Methods {
```
jasonpaulos, line 2250: It's also worthwhile to verify that the number of methods is the same on interface1 and 2.

Othewrise interface2 could have more extra methods that aren't checked by this function. Same thing for `checkEqualContracts`.

That being said, since this is a test, it would be fine to use reflection to check equality if you wanted to do that instead.

test/steps_test.go
```diff
@@ -2058,3 +2102,457 @@ func tealCheckDryrun(result string) error {
...
+func checkTxnCount(givenTxnCount int) error {
+	correctTxnCount := abiMethod.GetTxCountFromMethod()
+	if correctTxnCount != givenTxnCount {
+		return fmt.Errorf("txn count %d != %d", givenTxnCount, correctTxnCount)
+	}
+	return nil
+}
+
+func checkMethodSelector(givenMethodSelector string) error {
+	correctMethodSelector := fmt.Sprintf("%x", abiMethod.GetSelector())
```
jasonpaulos, line 2187: nit: `hex.DecodeString` or `hex.EncodeToString` could be used instead

test/steps_test.go
```diff
@@ -2058,3 +2102,457 @@ func tealCheckDryrun(result string) error {
...
+
+func checkSerializedMethodObject(jsonFile, loadedFrom string) error {
+	directory := path.Join("./features/unit/", loadedFrom)
+	jsons, err := loadMockJsons(jsonFile, directory)
+	if err != nil {
+		return err
+	}
+
+	correctJson := string(jsons[0])
+	if correctJson != abiJsonString {
```
jasonpaulos, line 2130: Note that this type of equality checking for JSON can be unnecessarily strict, since it requires the object field order to be the same.

Instead you could parse both JSON strings into an `interface{}` and then use reflection to check that they are deep equal.

But if it works for now, this can be addressed in a follow up PR when needed.

future/atomicTransactionComposer.go
```diff
@@ -0,0 +1,578 @@
...
+}
+
+func MakeAtomicTransactionComposer() AtomicTransactionComposer {
```
jasonpaulos, line 124: Since all  `AtomicTransactionComposer` fields have acceptable zero values, we could get rid of this method and instruct users to just use the zero value of the type, like this:

```go
var composer AtomicTransactionComposer
composer.AddTransaction(...)
composer.Execute(...)
```

What do you think?

jasonpaulos (commented):
This is nearly there, these are my final comments

abi/abi.go
ahangsu: I see, that makes very much sense.

jasonpaulos: To be clear, I am suggesting that `interactions.go` should move into the `abi` package @algoidurovic

future/atomicTransactionComposer.go
algoidurovic: The initial purpose of the caches was to avoid having to iterate over the txContexts and accumulate the extracted values. Maybe extraction/accumulation isn't expensive enough to justify this optimization?

future/atomicTransactionComposer.go
algoidurovic: I agree. I was originally expecting to have more complex logic for composer creation but that didn't turn out to be the case.

future/atomicTransactionComposer.go
jasonpaulos: > Maybe extraction/accumulation isn't expensive enough to justify this optimization?

That's exactly my thought. Especially since transaction composers can have at most 16 transactions, the cost of extracting these fields is pretty minor.

algoidurovic: Makes sense, caches seem like overkill here.

jasonpaulos (approved):
Looks good!

PR #262, opened 2021-11-18
verify state proof

shiqizng:
This PR adds a method to support state proof validation. meklearray module and relevant files are copied from go-algorand  to avoid importing libsodium which needs a C compiler.


shiqizng:
stateproof module is what I added. Everything else are vendor files or copied from go-algorand.

jasonpaulos (commented):
I _think_ the main purpose of this addition is to allow verifying the response from the `/v2/blocks/{round}/transactions/{txid}/proof` algod endpoint. The following suggestions are based on this:

1. It would probably be a lot more useful if the `stateproof.Verify` function had input types that match what the Go SDK already has available, not the types from go-algorand. Based on `ProofResponse` from `client/v2/common/models/proof_response.go`, a more useful definition might be:

```go
func Verify(root []byte, elemsToProve map[uint64][]byte, proof []byte, hashType string) error
```

2. If the `algod-crypto` package is meant to be temporary and should not be imported by consumers of this SDK (and assuming you do the first suggestion), it should probably go into the package `stateproof/internal`, which would effectively hide it from SDK consumers.

algoidan:
Hi everyone
I'm closing this PR for now for the following reasons:
1- we are still not sure if we need to support `VerifyMerklePath` on SDK
2- we need to discuss whether it is necessary to import all the Merkel code. we can easily implement a simple `VerifyMerklePath`

PR #264, opened 2021-11-21
adding EncodeAddress

### LINKED ISSUE #263: Add missing EncodeAddress
## Problem

At the moment we provide [DecodeAddress](https://github.com/algorand/go-algorand-sdk/blob/develop/types/address.go#L39) and you can get EncodeAddress behavior by taking a slice and copying it into the type of Address (32 byte array) then calling `.String()` on it but it is less than obvious.

## Solution

Implement EncodeAddress that does this for you and matches APIs of other SDKs

barnjamin:
Closes #263

### DIFF
--- a/types/address.go
+++ b/types/address.go
@@ -67,3 +67,18 @@ func DecodeAddress(addr string) (a Address, err error) {
 	copy(a[:], addressBytes)
 	return a, nil
 }
+
+// EncodeAddress turns a byte slice into the human readable representation of the address.
+// This representation includes a 4-byte checksum
+func EncodeAddress(addr []byte) (a string, err error) {
+	if len(addr) != hashLenBytes {
+		err = errWrongAddressByteLen
+		return
+	}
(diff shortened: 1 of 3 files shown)


jasonpaulos:
I like how this mirrors `DecodeAddress`, but it does not directly address the problem of making it easier to instantiate an `Address` object.

Would a function like `AddressFromBytes(b []byte) Address` be better?

Both are acceptable solutions to me, I just wanted to bring up the other option too.

pbennett:
@jasonpaulos EncodeAddress mirrors the js-algorand-sdk call by the same name.

https://github.com/algorand/js-algorand-sdk/blob/b2b5fff5d1752f9f4d59b28ddd1c64efaf33c51a/src/encoding/address.ts#L102

All the SDKs should be mirror each other wherever possible IMO.

jasonpaulos (approved):

PR #265, opened 2021-12-05
Adding CreateDryrun function

barnjamin:


future/dryrun.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	for _, acct := range accts {
+		acctInfo, err := client.AccountInformation(acct.String()).Do(context.Background())
+		if err != nil {
+			return drr, fmt.Errorf("failed to get application %s: %+v", acct, err)
+		}
+
+		drr.Accounts = append(drr.Accounts, acctInfo)
+	}
```
ahangsu, line 93: nit: I found in TS implementation that you de-duplicated the `acounts` and `apps` by maintaining a set. Is it practical in golang-implementation?

barnjamin: good point, maybe ill switch these to maps instead

future/dryrun.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		accts = append(accts, t.Txn.Sender)
+
+		accts = append(accts, t.Txn.Accounts...)
+		apps = append(apps, t.Txn.ForeignApps...)
+		assets = append(assets, t.Txn.ForeignAssets...)
+
+		if t.Txn.ApplicationID == 0 {
+			appId := 1 //TODO: make the magic number?
```
ahangsu, line 42: nit: I don't have a strong feeling on this one, probably need Andrew and Jason to weigh in on this.

ahangsu: okay, @jasonpaulos told me there is a `defaultAppIdx` in goal, which is here: https://github.com/algorand/go-algorand/blob/master/libgoal/libgoal.go#L1006.

It is applied in https://github.com/algorand/go-algorand/blob/master/libgoal/libgoal.go#L1066-L1090, which is `MakeDryrunStateGenerated`.

I think this applies to TS-implementation too? What do you think?

barnjamin: I can set it, I don't _think_ its required to be anything but !0 but for uniformity I'll make it match

ahangsu: TY!

future/dryrun.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/algorand/go-algorand-sdk/types"
+)
+
+const (
+	defaultAppId uint64 = 1380011588
+)
+
+// CreateDryrun creates a DryrunRequest object from a client and slice of SignedTxn objects and a default configuration
+// Passed in as a pointer to a DryrunRequest object to use for extra parameters
+func CreateDryrun(client *algod.Client, txns []types.SignedTxn, dr *models.DryrunRequest) (drr models.DryrunRequest, err error) {
```
jasonpaulos, line 19: Instead of using `context.Background()` in the body of this function, could you make the context an argument? That way the caller has control of it.

PR #267, opened 2021-12-07
Fix wait for confirmation function

jasonpaulos:
Fixes the following issues with the wait for confirmation function:
1. `PoolError` was being ignored
2. If the call to `PendingTransactionInformation` returns an error, that error should be ignored and the function should continue retrying if `waitRounds` allows it.
3. Removed the special value 0 for `waitRounds` which allows indefinitely waiting. This is a dangerous feature, since your code might actually wait forever if it never sees the transaction fail. Instead a value of 1000 (the maximum txn life) can be used to imply the same thing, and in reality is the longest you will ever have to wait.

### DIFF
--- a/future/waitForConfirmation.go
+++ b/future/waitForConfirmation.go
@@ -11,7 +11,7 @@ import (
 
 // `WaitForConfirmation` waits for a pending transaction to be accepted by the network
 // `txid`: The ID of the pending transaction to wait for
-// `waitRounds`: The number of rounds to block before existing with an error. If zero, there is no timeout
+// `waitRounds`: The number of rounds to block before exiting with an error.
 func WaitForConfirmation(c *algod.Client, txid string, waitRounds uint64, ctx context.Context, headers ...*common.Header) (txInfo models.PendingTransactionInfoResponse, err error) {
 	response, err := c.Status().Do(ctx, headers...)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


winder (approved):

future/waitForConfirmation.go
```diff
@@ -11,7 +11,7 @@ import (
 
 // `WaitForConfirmation` waits for a pending transaction to be accepted by the network
 // `txid`: The ID of the pending transaction to wait for
-// `waitRounds`: The number of rounds to block before existing with an error. If zero, there is no timeout
+// `waitRounds`: The number of rounds to block before exiting with an error.
 func WaitForConfirmation(c *algod.Client, txid string, waitRounds uint64, ctx context.Context, headers ...*common.Header) (txInfo models.PendingTransactionInfoResponse, err error) {
```
pbennett, line 15: I see this was already merged, but one thing I did with my version of this function and which I might still suggest if it's not too late, is using the Context for setting deadlines.  It's already designed for this in fact.  Don't specify specific blocks to wait for, but just use the passed-in context.

The for loop can then become
```
for {
	select {
		case <-ctx.Done():
			err = fmt.Errorf("Wait for transaction id %s timed out", txid)
			return
		default:
			break
	}
	
```
	

Callers can then use WaitForConfirmation(c, txid, ctx.WithTimeout(30 * time.Second)), or ctx.WithDeadline(...)
The key is users can chain existing Context's passed in from parent stacks, and honor explicit cancels.  Passed in context's should always be honored.  A top-level cancel shouldn't be ignored because a wait is using an independent mechanism.

If a default is needed the function could call ctx.Deadline() and if a deadline isn't set, construct a new Context copy with a default timeout.

jasonpaulos: The current implementation does use the passed in context as a time-based timeout, albeit indirectly, since it passes the context to the algod client when making requests. So if the context expires, then the algod client will return an error and `WaitForConfirmation` will terminate then. (There is one hole in this: errors from `PendingTransactionInformation` are ignored, but `StatusAfterBlock` is called immediately after and its errors aren't ignored, so there is no significant difference there.)

But block-based timeouts are also useful (and all other SDKs support them), so we kept the `waitRounds` parameter. Though if you want to only have a time-based timeout, then you could pass a very large number for this parameter, like 1000, to effectively ignore it.

I suppose would could create a custom context implementation that somehow has a block-based timeout, but that seems complicated.

pbennett: Yeah, good point on the socket connection getting canceled.  Block-based counts just worry me since the timing of blocks will change.  I suspect developers will code in a specific number of blocks based on a loose thinking on time.
For my version of this function I ended up using optional channel parameters as well so I could sometimes do it asynchronously if necessary.
Either way, it's a good add to the SDK to have a basic function for this already there.  It was a surprising omission.
👍

PR #273, opened 2021-12-21
Support ABI reference types and other improvements

### LINKED ISSUE #260: Support foreign objects as ABI arguments
## Summary
This is a follow up to #258, which added support for invoking ABI methods.

In addition to standard ABI type arguments, the SDK should support the special foreign object argument types `account`, `asset`, and `application`, [as described in ARC-4](https://github.com/algorandfoundation/ARCs/blob/6a9f91aed8068bbace872483a64cbf22d0e1c975/ARCs/arc-0004.md#foreign-apps-asas-and-accounts).

## Scope
Extend the `AtomicTransactionComposer.AddMethodCall` function to support these new method argument types. At a high level it should:

1. Recognize when an argument is a special type or a standard ABI type. The special argument types are only valid as top-level arguments, so direct string comparison should be sufficient.
2. If a special type is present, insert the argument value into the appropriate foreign array, and use the array index of the value as the app call argument, treated as an ABI `uint8`.

### LINKED ISSUE #268: Implement ABI changes
Implement the changes described in https://github.com/algorandfoundation/ARCs/pull/53

Additionally, the following implementation details should change:
* The atomic transaction composer should not add a group ID if there's only 1 transaction in the group, since it's unnecessary
* The atomic transaction composer should check that the type of a transaction argument (e.g. `pay`, `axfer`, etc.) matches the ABI transaction argument type.

jasonpaulos:
- [x] Add reference type support
- [x] Return error if transaction type argument does not match transaction argument type
- [x] AtomicTransactionComposer does not assign group ID if there's only 1 transaction in the group
- [x] AtomicTransactionComposer's `Execute` methods updates status to submitted before waiting for txn confirmation
- [x] Changes from https://github.com/algorandfoundation/ARCs/pull/53
  - [x] Don't tuple last argument when length is 15
  - [x] Modify `Contract` class to have `Networks` field
- [x] Add Contract and Interface descriptions, from  https://github.com/algorandfoundation/ARCs/pull/55
- [x] Changes from https://github.com/algorandfoundation/ARCs/pull/57
  - [x] AtomicTransactionComposer only checks the last log for a return value
  - [x] App creation and update transactions are supported in ABI method calls
- [x] Other minor changes
  - [x] Make comments use `//` and add more comments
  - [x] Add convenience functions `IsTransactionType` and `IsReferenceType`
  - [x] Add convenience methods to `Arg` and `Return` to check type information and get `abi.Type` value

Closes #268, closes #260 

Note: I wanted to update the go-algorand package to use additional ABI functions that were defined there, but I'm unable to update past this commit: https://github.com/algorand/go-algorand/commit/d3bbe62caa11a2608c318919122457d4512045c5, the reason being it references the package `github.com/algorand/go-algorand/data/basics` (which in turn references crypto packages with C dependencies) from `github.com/algorand/go-algorand/data/abi`, which breaks the current build. I will make a separate issue for this in go-algorand.

### DIFF
--- a/future/atomicTransactionComposer.go
+++ b/future/atomicTransactionComposer.go
@@ -8,77 +8,93 @@ import (
 
 	"github.com/algorand/go-algorand-sdk/abi"
 	"github.com/algorand/go-algorand-sdk/client/v2/algod"
+	"github.com/algorand/go-algorand-sdk/client/v2/common/models"
 	"github.com/algorand/go-algorand-sdk/crypto"
 	"github.com/algorand/go-algorand-sdk/types"
 )
 
-var ABI_RETURN_HASH = []byte{0x15, 0x1f, 0x7c, 0x75}
+// abiReturnHash is the 4-byte prefix for logged return values
+var abiReturnHash = []byte{0x15, 0x1f, 0x7c, 0x75}
 
-/** Represents an unsigned transactions and a signer that can authorize that transaction. */
+// TransactionWithSigner represents an unsigned transactions and a signer that can authorize that
+// transaction.
 type TransactionWithSigner struct {
-	/** An unsigned transaction */
+	// An unsigned transaction
 	Txn types.Transaction
-	/** A transaction signer that can authorize txn */
+	// A transaction signer that can authorize the transaction
 	Signer TransactionSigner
 }
 
-/** Represents the output from a successful ABI method call. */
+// Represents the output from a successful ABI method call.
 type ABIResult struct {
-	/** The TxID of the transaction that invoked the ABI method call. */
+	// The TxID of the transaction that invoked the ABI method call.
 	TxID string
-	/**
-	 * The raw bytes of the return value from the ABI method call. This will be empty if the method
-	 * does not return a value (return type "void").
-	 */
+	// The raw bytes of the return value from the ABI method call. This will be empty if the method
+	// does not return a value (return type "void").
 	RawReturnValue []byte
-	/**
-	 * The return value from the ABI method call. This will be undefined if the method does not return
-	 * a value (return type "void"), or if the SDK was unable to decode the returned value.
-	 */
+	// The return value from the ABI method call. This will be nil if the method does not return
+	// a value (return type "void"), or if the SDK was unable to decode the returned value.
 	ReturnValue interface{}
-	/** If the SDK was unable to decode a return value, the error will be here. */
+	// If the SDK was unable to decode a return value, the error will be here. Make sure to check
+	// this before examinging ReturnValue
 	DecodeError error
 }
 
+// AddMethodCallParams contains the parameters for the method AtomicTransactionComposer.AddMethodCall
 type AddMethodCallParams struct {
-	/** The ID of the smart contract to call */
+	// The ID of the smart contract to call. Set this to 0 to indicate an application creation call.
 	AppID uint64
-	/** The method to call on the smart contract */
+	// The method to call on the smart contract
 	Method abi.Method
-	/** The arguments to include in the method call. If omitted, no arguments will be passed to the method. */
+	// The arguments to include in the method call. If omitted, no arguments will be passed to the
+	// method.
 	MethodArgs []interface{}
-	/** The address of the sender of this application call */
+	// The address of the sender of this application call
 	Sender types.Address
-	/** Transactions params to use for this application call */
+	// Transactions params to use for this application call
 	SuggestedParams types.SuggestedParams
-	/** The OnComplete action to take for this application call. If omitted, OnApplicationComplete.NoOpOC will be used. */
+	// The OnComplete action to take for this application call
 	OnComplete types.OnCompletion
-	/** The note value for this application call */
+	// The approval program for this application call. Only set this if this is an application
+	// creation call, or if onComplete is UpdateApplicationOC.
+	ApprovalProgram []byte
+	// The clear program for this application call. Only set this if this is an application creation
+	// call, or if onComplete is UpdateApplicationOC.
+	ClearProgram []byte
+	// The global schema sizes. Only set this if this is an application creation call.
+	GlobalSchema types.StateSchema
+	// The local schema sizes. Only set this if this is an application creation call.
+	LocalSchema types.StateSchema
+	// The number of extra pages to allocate for the application's programs. Only set this if this
+	// is an application creation call.
+	ExtraPages uint32
+	// The note value for this application call
 	Note []byte
-	/** The lease value for this application call */
+	// The lease value for this application call
 	Lease [32]byte
-	/** If provided, the address that the sender will be rekeyed to at the conclusion of this application call */
+	// If provided, the address that the sender will be rekeyed to at the conclusion of this application call
 	RekeyTo types.Address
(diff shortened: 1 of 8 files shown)


future/atomicTransactionComposer.go
```diff
@@ -212,56 +221,104 @@ func (atc *AtomicTransactionComposer) AddMethodCall(params AddMethodCallParams)
...
 			return err
 		}
 
-		abiArgs = abiArgs[:14]
-		abiArgs = append(abiArgs, tupleValues)
-		abiTypes = abiTypes[:14]
-		abiTypes = append(abiTypes, tupleType)
+		basicArgValues = append(basicArgValues[:14], valueForTuple)
+		basicArgTypes = append(basicArgTypes[:14], tupleType)
 	}
```
jasonpaulos, line 316: The cucumber tests exposed a bug in this section of code. The problem was this part:

```go
tupleValues := abiArgs[14:]
// ...
abiArgs = abiArgs[:14]
abiArgs = append(abiArgs, tupleValues)
```

which is an issue because `abiArgs[14:]` and `abiArgs[:14]` are backed by the same array, and `append(abiArgs[:14], abiArgs[14:])` does not really append to `abiArgs[:14]`, it sets index 15 to `abiArgs[14:]`, **which contains index 15**. So I believe you end up with the first element of `tupleValues` being a reference to itself.

The solution was to make a copy of the tuple part of the array so that it does not get modified when it's appended back into the original array.

future/atomicTransactionComposer.go
```diff
@@ -494,65 +548,204 @@ func (atc *AtomicTransactionComposer) Execute(client *algod.Client, ctx context.
...
+}
 
-				abiValue, err := abiType.Decode(log[4:])
-				if err != nil {
-					returnValues = append(returnValues, ABIResult{TxID: txid, RawReturnValue: log[4:], ReturnValue: nil, DecodeError: err})
+// populateMethodCallReferenceArgs parses reference argument types and resolves them to an index
+// into the appropriate foreign array. Their placement will be as compact as possible, which means
+// values will be deduplicated and any value that is the sender or the current app will not be added
+// to the foreign array.
+func populateMethodCallReferenceArgs(sender string, currentApp uint64, types []string, values []interface{}, accounts *[]string, apps *[]uint64, assets *[]uint64) ([]int, error) {
```
jasonpaulos, line 677: This function is based on https://github.com/algorand/go-algorand/blob/6be3ffa4d2a0499b3c407475338327e758e22132/cmd/goal/application.go#L1091

future/atomicTransactionComposer.go
jasonpaulos: I believe goal's implementation suffers from this as well, and Python and Java may also be vulnerable to it, but I haven't checked.

future/atomicTransactionComposer.go
```diff
@@ -212,56 +221,104 @@ func (atc *AtomicTransactionComposer) AddMethodCall(params AddMethodCallParams)
...
+	// Up to 16 app arguments can be passed to app call. First is reserved for method selector,
+	// and the rest are for method call arguments. But if more than 15 method call arguments
+	// are present, then the 14th+ are placed in a tuple in the last app argument slot
+	if len(basicArgValues) > 15 {
+		typesForTuple := make([]abi.Type, len(basicArgTypes)-14)
+		copy(typesForTuple, basicArgTypes[14:])
+
+		valueForTuple := make([]interface{}, len(basicArgValues)-14)
+		copy(valueForTuple, basicArgValues[14:])
+
```
ahangsu, line 308: nit: better use some constant like `WRAP_THRESHOLD = 14` rather than some magic constant

future/atomicTransactionComposer.go
ahangsu: oof, it is such a weird thing that the index 15 (count from 0) is a reference to itself after appending, nice catch.
I don't think it happens in Java, I handled it in another way.
I also don't think it will happen in Python, can be verified with some lines of code:
```py
stuff = [i for i in range(20)]
stuff_tail = stuff[14:]
stuff = stuff[:14]
stuff.append(stuff_tail)
print(stuff)
```

future/atomicTransactionComposer.go
ahangsu: Hoping for generic be introduced to simplify the code 🙏

future/atomicTransactionComposer.go
```diff
@@ -494,65 +548,204 @@ func (atc *AtomicTransactionComposer) Execute(client *algod.Client, ctx context.
...
+		return "", err
+	}
+	encoded, err := abiType.Encode(value)
+	if err != nil {
+		return "", err
+	}
+	decoded, err := abiType.Decode(encoded)
+	if err != nil {
+		return "", err
+	}
```
ahangsu, line 663: question: encode/decode to check address byte validity?

future/atomicTransactionComposer.go
```diff
@@ -494,65 +548,204 @@ func (atc *AtomicTransactionComposer) Execute(client *algod.Client, ctx context.
...
+	indexToWaitFor := 0
+	for i, txContext := range atc.txContexts {
+		if txContext.isMethodCallTx() {
+			// if there is a method call in the group, we need to query the
+			// pending tranaction endpoint for it anyway, so as an optimization
+			// we should wait for its TxID
+			indexToWaitFor = i
+			break
+		}
+	}
```
ahangsu, line 562: good optimization.

future/atomicTransactionComposer.go
```diff
@@ -212,56 +221,104 @@ func (atc *AtomicTransactionComposer) AddMethodCall(params AddMethodCallParams)
...
 
-	// Up to 16 args can be passed to app call. First is reserved for method selector and last is
-	// reserved for remaining args packaged into one tuple.
-	if len(abiArgs) > 14 {
-		tupleTypes := abiTypes[14:]
-		tupleValues := abiArgs[14:]
-		tupleType, err := abi.MakeTupleType(tupleTypes)
+	var foreignAccounts []string
+	var foreignApps []uint64
+	var foreignAssets []uint64
```
ahangsu, line 288: same here: do we need to constrain on the number of foreign objects (as discussed in JS-sdk)

ahangsu (commented):
Looks largely good to me.

future/atomicTransactionComposer.go
```diff
@@ -212,56 +228,104 @@ func (atc *AtomicTransactionComposer) AddMethodCall(params AddMethodCallParams)
...
+	}
+	for i, resolved := range refArgsResolved {
+		basicArgIndex := refArgIndexToBasicArgIndex[i]
+		// use the foreign array index as the encoded argument value
+		basicArgValues[basicArgIndex] = resolved
+	}
+
+	// Up to 16 app arguments can be passed to app call. First is reserved for method selector,
+	// and the rest are for method call arguments. But if more than 15 method call arguments
+	// are present, then the 14th+ are placed in a tuple in the last app argument slot
```
ahangsu, line 308:
```suggestion
	// are present, then the method arguments after the 14th are placed in a tuple in the last app argument slot
```
(same as the `goal method` PR)

ahangsu (approved):
LGTM

PR #276, opened 2022-01-06
Create response object for `AtomicTransactionComposer.Execute`

jasonpaulos:
Instead of having `AtomicTransactionComposer.Execute` return a tuple of many values, this PR creates an object, `ExecuteResult`, which holds the values being returned.

This will allow us to potentially add more fields in the future without changing the method signature.

Also, rename `ABIResult` to `ABIMethodResult` for clarity.

### DIFF
--- a/future/atomicTransactionComposer.go
+++ b/future/atomicTransactionComposer.go
@@ -33,8 +33,8 @@ type TransactionWithSigner struct {
 	Signer TransactionSigner
 }
 
-// Represents the output from a successful ABI method call.
-type ABIResult struct {
+// ABIMethodResult represents the output from a successful ABI method call.
+type ABIMethodResult struct {
 	// The TxID of the transaction that invoked the ABI method call.
 	TxID string
 	// The raw bytes of the return value from the ABI method call. This will be empty if the method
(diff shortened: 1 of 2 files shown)


future/atomicTransactionComposer.go
```diff
@@ -572,19 +584,23 @@ func (atc *AtomicTransactionComposer) Execute(client *algod.Client, ctx context.
...
-		return 0, nil, nil, err
+		return ExecuteResult{}, err
 	}
 	atc.status = COMMITTED
 
-	var returnValues []ABIResult
+	executeResponse := ExecuteResult{
+		ConfirmedRound: txinfo.ConfirmedRound,
+		TxIDs:          atc.getTxIDs(),
+	}
```
ahangsu, line 594: nit: okay to initialize the `MethodResults` slice with size of `TxIDs`, so no need for append later in loop?
i.e., `make([]ABIMethodResults, len(atc.getTxIDs()))` (just an example)

ahangsu (commented):
Looks largely good to me, waiting on other folks to say something.

future/atomicTransactionComposer.go
```diff
@@ -86,6 +86,18 @@ type AddMethodCallParams struct {
 	Signer TransactionSigner
 }
 
+// ExecuteResult contains the results of successfully calling the Execute method on an
+// AtomicTransactionComposer object.
+type ExecuteResult struct {
+	// The round in which the executed transaction group with confirmed on chain
```
tzaffi, line 92:
```suggestion
	// The round in which the executed transaction group was confirmed on chain
```

tzaffi (approved):
LGTM with one grammatical suggestion

algochoi (approved):

ahangsu (approved):
LGTM

PR #284, opened 2022-01-19
Stateproof keys APIs changes

algoidan:


jasonpaulos (commented):
Mostly looks good. Though once stateproofs are merged into master, I would like to do a quick review again

algoidan:
We are not going to enable the compact cert in this PR. If I understand the problem correctly, this issue will be resolved once the SDK will support the compact certs (which they don't right now)

winder (approved):
I was thrown by the missing cucumber definition for `default V2 key registration transaction {string}`, but see that it was added in an earlier PR.

PR #286, opened 2022-01-31
adding status code checker to msgpack decode

### LINKED ISSUE #285: msgpack decode masks http error
### Subject of the issue

When calling something like  `client.Block(round).Do...`, if there is an HTTP error, JSON is returned but we don't check status code and try to msgpack decode the result anyway.  

This leads to the error _seeming_ like its one with formatting and the real error is masked. 

We should check the http status code here and return an error that wraps the resulting error message https://github.com/algorand/go-algorand-sdk/blob/7285700bfe9d475535b28fbaa0f770caf3602000/client/v2/common/common.go#L220

barnjamin:
closes #285


ahangsu (approved):
I don't know that if HTTP error then we get JSON back rather than msgpack, but these lines of codes seems to be making sense to me, which essentially check the body and handle the error message return under different HTTP status code.

barnjamin:
>I don't know that if HTTP error then we get JSON back rather than msgpack, but these lines of codes seems to be making sense to me, which essentially check the body and handle the error message return under different HTTP status code.

Only reason I noticed this is because I did get JSON back and was v confused by the error message 😄

client/v2/common/common.go
```diff
@@ -220,6 +220,16 @@ func (client *Client) GetRawMsgpack(ctx context.Context, response interface{}, p
 
 	defer resp.Body.Close()
 
+	if resp.StatusCode != 200 {
```
norunners, line 223: Prefer `http.StatusOK` over `200`.

barnjamin: good point

ahangsu:
@barnjamin you might also want to merge #287, we are moving to circle-ci sooon...

jasonpaulos (approved):
Looks good from my point of view

barnjamin:
Am I allowed to just merge this?

ahangsu:
I think so, since both @jasonpaulos and I felt good about it...

PR #296, opened 2022-02-23
Readme updates

barnjamin:


algochoi (approved):
I think this is a good change. The existing examples could be outdated and may confuse new people - I'm also in favor of doing this in the Java SDK as well.

PR #301, opened 2022-03-22
update go version to 1.16

tsachiherman:
## Issue

update go version to 1.16 in the go.mod file, as well all the dependencies ( which now also use go 1.16 ).


tsachiherman:
> Why are you vendoring some of the libraries?

It's a good question, but I'm not the right person to ask.. We were *already* vendoring these packages. I just updated these.. let me check that again, as it seems to fail the test.

winder:
> > Why are you vendoring some of the libraries?
> 
> It's a good question, but I'm not the right person to ask.. We were _already_ vendoring these packages. I just updated these.. let me check that again, as it seems to fail the test.

Gotcha. Many of the files are new, but it sounds like they are just new to the library after updating versions.

Issue #305, opened 2022-03-27
Missing ParticipationUpdates struct


barnjamin:
### Subject of the issue

When decoding a block, if the struct being decoded into does not have all the fields defined the decoder will error out. We're missing at least the ParticipationUpdates struct that is embedded in the BlockHeader

In go-algorand-sdk:
https://github.com/algorand/go-algorand-sdk/blob/3f2d85658f808232fcd8d4d4303a74f7c6171421/types/block.go#L96

In go-algorand
https://github.com/algorand/go-algorand/blob/master/data/bookkeeping/block.go#L142

pbennett:
Thanks for this - this is a critical nasty one.

winder:
Closed by #306

PR #306, opened 2022-03-27
adding ParticipationUpdates and CompactCertState to BlockHeader

barnjamin:
Addresses #305

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -93,11 +93,6 @@ type (
 		// started being supported).
 		TxnCounter uint64 `codec:"tc"`
 
-		// CompactCert tracks the state of compact certs, potentially
-		// for multiple types of certs.
-		//msgp:sort protocol.CompactCertType protocol.SortCompactCertType
-		//CompactCert map[protocol.CompactCertType]CompactCertState `codec:"cc"`
-
 		// ParticipationUpdates contains the information needed to mark
 		// certain accounts offline because their participation keys expired
 		ParticipationUpdates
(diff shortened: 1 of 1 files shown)


barnjamin:
This should probably have its own test that somehow also fails if go-algorand changes?

Can we just use the structs defined in go-algorand for this?

pbennett:
Thanks for this fix - this has happened before as well - the duplication of these core structures either needs unified, or there needs to be good tests to catch missing additions.
This caused a halt on my block watcher on testnet which has me concerned over what other sdks might have the same issue and the impact on mainnet for integrators.

types/block.go
```diff
@@ -152,6 +172,29 @@ type (
 		NextProtocolSwitchOn   Round  `codec:"nextswitch"`
 	}
 
+	// CompactCertState tracks the state of compact certificates.
+	CompactCertState struct {
```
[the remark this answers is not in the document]
barnjamin: fair enough, merged Wills suggestion

types/block.go
```diff
@@ -92,6 +92,26 @@ type (
 		// transactions have ever been committed (since TxnCounter
 		// started being supported).
 		TxnCounter uint64 `codec:"tc"`
+
+		// CompactCert tracks the state of compact certs, potentially
+		// for multiple types of certs.
+		//msgp:sort protocol.CompactCertType protocol.SortCompactCertType
+		//CompactCert map[protocol.CompactCertType]CompactCertState `codec:"cc"`
```
algoidan, line 99: nits: should we remove this comment as well?

barnjamin: Yep, I'll push this change as well

algoidan (approved):

winder (approved):

PR #307, opened 2022-03-28
Avoid client response failure on unknown field.

winder:
Change the codec parser to allow unknown fields when processing msgpack objects returned by the REST API.

### DIFF
--- a/encoding/msgpack/msgpack.go
+++ b/encoding/msgpack/msgpack.go
@@ -10,6 +10,9 @@ import (
 // with our settings (canonical, paranoid about decoding errors)
 var CodecHandle *codec.MsgpackHandle
 
+// LenientCodecHandle is used to instantiate msgpack encoders for the REST API.
+var LenientCodecHandle *codec.MsgpackHandle
+
 // init configures our msgpack encoder and decoder
 func init() {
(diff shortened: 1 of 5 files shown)


barnjamin (approved):

encoding/msgpack/msgpack.go
```diff
@@ -19,6 +22,15 @@ func init() {
 	CodecHandle.RecursiveEmptyCheck = true
 	CodecHandle.WriteExt = true
 	CodecHandle.PositiveIntUnsigned = true
+
+	LenientCodecHandle = new(codec.MsgpackHandle)
+	// allow unknown fields to ensure forward compatibility.
+	LenientCodecHandle.ErrorIfNoField = false
```
shiqizng, line 28: Why do we want allow for unmatched field instead of updating the struct with field changes?

winder: The main case is probably when someone did not update the SDK. Suppose there are new fields in the response, we have the following options:
1) allow the SDK to continue working by ignoring the missing field.
2) force the application to update the SDK to a new version which has the missing field.

winder: We should also ensure the types are updated / correct.

shiqizng (approved):

PR #309, opened 2022-04-10
Added support for SHA256 TxnRoot header

Aharonee:


algoidan:
done on https://github.com/algorand/go-algorand-sdk/pull/374

PR #314, opened 2022-05-03
build: Bump golang to 1.17

Eric-Warehime:
Bumping golang version to 1.17 to bring the go-sdk into parity with the go-algorand repository update happening in https://github.com/algorand/go-algorand/pull/3919

build/test works and I've updated and run the docker tests as well.


Eric-Warehime:
Yeah I looked into the langspec change and it seems like for some reason the updates don't like the line length of the previous commit. So each line is shortened by one entry, but they're functionally the same. It just gets regenerated like that during build if you checkout HEAD so I just included it.

jasonpaulos:
Alternatively we could opt to use `go:embed` at this point instead of generating that file.

winder (approved):

PR #316, opened 2022-05-11
Generate updated API client code

algoidurovic:
Depends on: https://github.com/algorand/go-algorand/pull/3957


client/v2/algod/getProof.go
```diff
@@ -13,6 +13,11 @@ type GetProofParams struct {
 
 	// Format configures whether the response object is JSON or MessagePack encoded.
 	Format string `url:"format,omitempty"`
+
+	// Hashtype the type of hash function used to create the proof, must be one of:
+	// * sha512_256
+	// * sha256
+	Hashtype string `url:"hashtype,omitempty"`
```
jasonpaulos, line 20: I'm not sure if changes to this file and `client/v2/common/models/proof_response.go` are ready for release, and it looks like they'll be addressed in #309 anyway. So I'd suggest removing them from this PR for now.

client/v2/algod/tealDisassemble.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/algorand/go-algorand-sdk/client/v2/common/models"
+)
+
+// TealDisassemble given the base64 encoded program bytes, return the TEAL source
+// code in plain text. This endpoint is only enabled when a node's configuration
+// file sets EnableDeveloperAPI to true.
+type TealDisassemble struct {
+	c *Client
+
+	source string
```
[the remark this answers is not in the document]
jasonpaulos: Yes, there are two things wrong:
- The description should not mention base64, that's a problem with go-algorand: https://github.com/algorand/go-algorand/blob/aa82b9a6e78dd38c8115120ca90b817f6be73ef3/daemon/algod/api/algod.oas2.json#L1441
- This field should be `[]byte` since it's declared with `"format": "byte"` in the API spec. There may be an issue with the generator since that's not the case.

Since the TEAL disassembly has nothing to do with this PR, I'd suggest manually removing all references to it from the generated files and we can address them later.

client/v2/algod/tealDisassemble.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+// Do performs the HTTP request
+func (s *TealDisassemble) Do(ctx context.Context, headers ...*common.Header) (response models.DisassembleResponse, err error) {
+	err = s.c.get(ctx, &response, "/v2/teal/disassemble", nil, headers)
```
winder, line 21: According to the link jason shared, shouldn't this be a POST and include the bytes body?
https://github.com/algorand/go-algorand/blob/aa82b9a6e78dd38c8115120ca90b817f6be73ef3/daemon/algod/api/algod.oas2.json#L1441

algoidurovic: Sorry, this is a stale change. I pulled develop into this branch and now it's gone.

michaeldiamant (approved):

jasonpaulos (approved):

PR #320, opened 2022-05-13
Refresh generated sources

michaeldiamant:
Regenerates source using https://github.com/algorand/generator/pull/30 to fix the data type issue identified in https://github.com/algorand/go-algorand-sdk/pull/316#discussion_r872584788.

Diff generated via the following command from _generator_ home: `sh scripts/generate_go.sh`.

PR #321, opened 2022-05-20
Update generated files

Eric-Warehime:
Updated go-sdk using the following:   
https://github.com/algorand/generator/pull/31
https://github.com/algorand/indexer/pull/1005
 as well as the most recent algod spec file defined in go-algorand.


michaeldiamant (approved):
@Eric-Warehime Looks correct based on my comparison with https://github.com/algorand/go-algorand-sdk/pull/320.  The changes are equivalent suggesting no regressions introduced via https://github.com/algorand/generator/pull/31.

PR #323, opened 2022-05-25
Copy foreign arrays before modifying

algoidurovic:
Minor change to clarify pointer logic.


future/atomicTransactionComposer_test.go
```diff
@@ -194,6 +194,8 @@ func TestAddMethodCallWithManualForeignArgs(t *testing.T) {
 	require.Equal(t, len(txns[0].Txn.ForeignApps), 2)
 	require.Equal(t, txns[0].Txn.ForeignApps[0], types.AppIndex(1))
 	require.Equal(t, txns[0].Txn.ForeignApps[1], types.AppIndex(2))
+	// verify original params object hasn't changed.
+	require.Equal(t, len(params.ForeignApps), 1)
```
jasonpaulos, line 198: nit: might as well do `require.Equal(t, params.ForeignApps, []uint64{1})` to compare the value as well

jasonpaulos (approved):
Pending tests, looks good to me. Just a single optional nit

PR #324, opened 2022-05-25
Build: Sdk code generation automation

Eric-Warehime:
## Summary

Adds a github workflow which will call our code gen action.  
Depends on action being defined in  our generator... See https://github.com/algorand/generator/pull/32 for more details and testing done.

Runs on a cron job daily at 23:20. The Github docs say to try to avoid common times because there will be natural spikes of event processing happening then, so I've defined all of the jobs at an arbitrary time during the night.

It will run the code generator, commit to a branch if necessary, and open a PR for us to review. It will no-op if the branch w/ identical changes already exist.

### DIFF
--- a/.github/workflows/codegen.yml
+++ b/.github/workflows/codegen.yml
@@ -1,7 +1,7 @@
 name: "SDK Code Generation"
 on:
   schedule:
-    - cron: '*/5 * * * *'
+    - cron: '20 23 * * *'
 permissions:
   contents: write
   pull-requests: write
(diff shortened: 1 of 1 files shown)


.github/workflows/codegen.yml
```diff
@@ -0,0 +1,19 @@
...
+  contents: write
+  pull-requests: write
+jobs:
+  generate_and_pr:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+      - name: Generate and PR
+        uses: algorand/generator/.github/actions/sdk-codegen/
```
cce, line 15: why hide the action's code in a dot-directory?

algojack (approved):

.github/workflows/codegen.yml
Eric-Warehime: I just wanted to keep all github related things together in one directory. I could split off everything into a base level directory. 

I think it probably also makes sense to add a Readme to the action directory giving more context on the pieces, so there's some documentation around the whole setup missing probably.

winder (approved):
LGTM, lets see if it works.

PR #335, opened 2022-06-23
Dev Tools: Source map decoder

barnjamin:


logic/source_map.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	return sm
+}
+
+func (s *SourceMap) GetLineForPc(pc int) int {
+	return s.PcToLine[pc]
+}
+
+func (s *SourceMap) GetPCsForLine(line int) []int {
```
michaeldiamant, line 78: @barnjamin Minor and optional:  Elsewhere, the PR does _not_ capitalize _PC_.  Propose lowercase for consistency.

```suggestion
func (s *SourceMap) GetPcsForLine(line int) []int {
```

test/steps_test.go
```diff
@@ -329,6 +331,8 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^I get the method from the Interface by name "([^"]*)"$`, iGetTheMethodFromTheInterfaceByName)
 	s.Step(`^I get the method from the Contract by name "([^"]*)"$`, iGetTheMethodFromTheContractByName)
 	s.Step(`^the produced method signature should equal "([^"]*)"\. If there is an error it begins with "([^"]*)"$`, theProducedMethodSignatureShouldEqualIfThereIsAnErrorItBeginsWith)
+	s.Step(`^a source map json file "([^"]*)"$`, aSourceMapJsonFile)
+	s.Step(`^the string composed of pc:line number equals "([^"]*)"$`, theStringComposedOfPclineNumberEquals)
```
algoidurovic, line 335: Is there a reason why these tests are in @unit and use a mock server response? Seems more robust to have an actual server call in @integration in case the format of the response object changes.

logic/source_map.go
```diff
@@ -0,0 +1,136 @@
...
+
+const (
+	// consts used for vlq encoding/decoding
+	b64table     string = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
+	vlqShiftSize        = 5
+	vlqFlag             = 1 << vlqShiftSize
+	vlqMask             = vlqFlag - 1
+)
+
+func DecodeSourceMapLine(vlq string) []int {
```
algoidurovic, line 106: Seems like this is an helper function? Does it need to be exported?

algoidurovic, line 106: Optional: some unit tests can be copied from go-algorand to test edge cases here. Although, if the code is nearly the exact same it's not strictly necessary imo.

algoidurovic (commented):
Looks good, just some minor feedback.

test/steps_test.go
```diff
@@ -2616,3 +2620,32 @@ func callingAppTraceProduces(arg1 string) error {
...
+	if err := json.Unmarshal(b, &ism); err != nil {
+		return err
+	}
+
+	sourceMap, err = logic.DecodeSourceMap(ism)
+
+	return err
+}
+
+func theStringComposedOfPclineNumberEquals(arg1 string) error {
```
michaeldiamant, line 2640: @barnjamin Minor and optional:  Unless you prefer as is, can parameters like `arg1` be named according to intended use case?

test/steps_test.go
michaeldiamant: A few thoughts here:
* I'm on-board with an integration test primarily as a means to vet all SDKs can request a sourcemap during compilation.
* I think the existing unit test adds a valuable proof point and I'd optionally favor retaining it.  
* I agree there's value in defending against response changes.  However, since we're discussing an external protocol (sourcemap) that's remained stable for _years_,  I think the chance of response format changes are _low_.
  * Aside:  In future iterations where we also support PyTeal-to-TEAL mapping, I envision supporting multiple sources.  We'll need a richer SDK model + more tests.

client/v2/algod/tealCompile.go
```diff
@@ -36,6 +36,10 @@ func (s *TealCompile) Sourcemap(Sourcemap bool) *TealCompile {
 
 // Do performs the HTTP request
 func (s *TealCompile) Do(ctx context.Context, headers ...*common.Header) (response models.CompileResponse, err error) {
-	err = s.c.post(ctx, &response, "/v2/teal/compile", s.source, headers)
+	if s.p.Sourcemap {
```
michaeldiamant, line 39: @barnjamin Leaving some thoughts on how to remedy the broken query string parameter handling.  Let me know what you think?  Happy to discuss live.

First - Thanks for identifying + providing a fix for the problem.

I think my objectives for the PR are:
* Make the change _generator_-compatible.  Re-running the generator should _not_ cause a diff.
* Avoid changes impacting all routes.

Given those objectives, I see 2 paths forward:
* The path shown in the PR - Augment query string by-hand.
* Add parallel methods only used by the TEAL compile route.  I'm imagining a parallel `submitFormRaw` + upstream methods to pass query string parameters as a dedicated method argument.

A few thoughts:
* Either route requires updates to https://github.com/algorand/generator/blob/master/go_templates/query.vm#L239-L243.
* Philosophically, I prefer adding parallel methods to isolate impact + start moving towards a more robust solution.  In practice, I see it'd be significant code duplication.  I'd like to sleep on it, but I'm leaning towards taking the PR as is + _generator_ template modifications.

I'll likely add a go-algorand-sdk ticket to suggest providing a 1st class repo-wide solution.

barnjamin: I just tweaked this to be a little more `future-proof` by encoding the get parameters with the existing `go-querystring` dependency.  This pattern _should_ allow us to add further parameters as needed to the struct alone and have them encoded without touching the URL generation logic

logic/source_map.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+
+		sm.LineToPc[lastLine] = append(sm.LineToPc[lastLine], idx)
+		sm.PcToLine[idx] = lastLine
+	}
+
+	return sm, nil
+}
+
+func (s *SourceMap) GetLineForPc(pc int) int {
```
michaeldiamant, line 86: @barnjamin Friendly visual reminder to implement an analog of https://github.com/algorand/py-algorand-sdk/pull/353#discussion_r910177370 here.

client/v2/algod/tealCompile.go
michaeldiamant: @barnjamin https://github.com/algorand/go-algorand-sdk/pull/335/commits/da90c91574cf8d33d15389db0a64315c7162daa4 works for me provided _generator_ can be modified.  I'll leave the thread open until a pass at _generator_ is made.

test/steps_test.go
algoidurovic: I can get behind that reasoning. Resolving.

logic/source_map.go
michaeldiamant: @barnjamin You're correct that the logic won't panic.  When I wrote my earlier note, I forgot Go's behavior is to return the zero value for a non-existent key (https://go.dev/tour/moretypes/22).

My updated thinking is:
* `GetPcsForLine` requires no change because the zero value is empty array.
* `GetLineForPc` _should_ be changed to disambiguate the zero value.  0 can mean a valid mapping or the value returning when key does not exist.
  * I'm _assuming_ the preferred Go convention here is to return `int, bool` so that the caller can validate key presence.

client/v2/algod/tealCompile.go
barnjamin: https://github.com/algorand/generator/pull/37 PR'd

logic/source_map.go
michaeldiamant: @barnjamin Thanks for the change - resolving.

client/v2/algod/tealCompile.go
michaeldiamant: @barnjamin Resolving - The _generator_ change looks good.  Additionally, I opened an issue to document the underlying gap:  https://github.com/algorand/generator/issues/38.

michaeldiamant (approved):
@barnjamin Leaving approval with these assumptions prior to merge:
* https://github.com/algorand/generator/pull/37 is merged.
* https://github.com/algorand/algorand-sdk-testing/pull/199 is merged.
* `test/docker/run_docker.sh` branch change is reverted.

client/v2/common/common.go
```diff
@@ -103,27 +103,27 @@ func mergeRawQueries(q1, q2 string) string {
 }
 
 // submitFormRaw is a helper used for submitting (ex.) GETs and POSTs to the server
-func (client *Client) submitFormRaw(ctx context.Context, path string, body interface{}, requestMethod string, encodeJSON bool, headers []*Header) (resp *http.Response, err error) {
+func (client *Client) submitFormRaw(ctx context.Context, path string, body interface{}, requestMethod string, encodeJSON bool, headers []*Header, getParams interface{}) (resp *http.Response, err error) {
```
winder, line 106: What do you think about renaming body to params, and having the new parameter be body?

That way you don't have this weird conditional on the requestMethod to decide how to handle parameters.

client/v2/common/common.go
```diff
@@ -103,33 +103,34 @@ func mergeRawQueries(q1, q2 string) string {
...
+	var v url.Values
+
+	if params != nil {
+		v, err = query.Values(params)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if requestMethod == "POST" && rawRequestPaths[path] {
```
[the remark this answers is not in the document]
michaeldiamant: Resolving - After a brief verbal with @barnjamin, I convinced myself the PR's behavior is more appropriate.

michaeldiamant:
Merging - Based on a verbal with @winder, we agreed I can merge provided I'm comfortable with the diff.

PR #337, opened 2022-06-30
Support for Box array transaction field

michaeldiamant:
Supports supplying Box references during transaction submission.

### DIFF
--- a/client/v2/common/models/box.go
+++ b/client/v2/common/models/box.go
@@ -1,0 +1,10 @@
+package models
+
+// Box box name and its content.
+type Box struct {
+	// Name (name) box name, base64 encoded
+	Name []byte `json:"name"`
+
+	// Value (value) box value, base64 encoded.
+	Value []byte `json:"value"`
+}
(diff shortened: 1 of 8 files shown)


michaeldiamant:
The PR extracts _only_ Box references from #322 in an attempt to isolate change sets.

client/v2/common/models/box.go
```diff
@@ -0,0 +1,10 @@
+package models
```
algoidurovic, line 1: I believe this is code generated for the Box endpoint, so it should be in the other PR.

algoidurovic (commented):
Everything looks good besides the one detail mentioned below.

client/v2/common/models/box.go
michaeldiamant: @algoidurovic Thanks for catching - addressed via https://github.com/algorand/go-algorand-sdk/pull/337/commits/309afb3e783ab799085c3037e6c5c406f808ed92.

algoidurovic (approved):

PR #341, opened 2022-07-06
Boxes: Merge feature/box-storage into develop

michaeldiamant:
Provides a draft PR to simplify tracking changes to feature/box-storage.

### DIFF
--- a/future/transaction.go
+++ b/future/transaction.go
@@ -605,6 +605,9 @@ func byte64FromBase64(in string) (out [64]byte, err error) {
 // - extraPages    ExtraProgramPages specifies the additional app program size requested in pages.
 //                 A page is 1024 bytes. This field enables execution of app programs
 //                 larger than the default maximum program size.
+//
+// - boxes         lists the boxes to be accessed during evaluation of the application
+//                 call. This also must include the boxes accessed by inner app calls.
 
 // MakeApplicationCreateTx makes a transaction for creating an application (see above for args desc.)
 // - optIn: true for opting in on complete, false for no-op.
@@ -618,6 +621,7 @@ func MakeApplicationCreateTx(
 	accounts []string,
 	foreignApps []uint64,
 	foreignAssets []uint64,
+	appBoxReferences []types.AppBoxReference,
 	sp types.SuggestedParams,
 	sender types.Address,
 	note []byte,
@@ -636,6 +640,7 @@ func MakeApplicationCreateTx(
 		accounts,
 		foreignApps,
 		foreignAssets,
+		appBoxReferences,
 		oncomp,
 		approvalProg,
 		clearProg,
@@ -660,6 +665,7 @@ func MakeApplicationCreateTxWithExtraPages(
 	accounts []string,
 	foreignApps []uint64,
 	foreignAssets []uint64,
+	appBoxReferences []types.AppBoxReference,
 	sp types.SuggestedParams,
 	sender types.Address,
 	note []byte,
@@ -678,6 +684,7 @@ func MakeApplicationCreateTxWithExtraPages(
 		accounts,
 		foreignApps,
 		foreignAssets,
+		appBoxReferences,
 		oncomp,
 		approvalProg,
 		clearProg,
@@ -703,6 +710,7 @@ func MakeApplicationUpdateTx(
 	accounts []string,
 	foreignApps []uint64,
 	foreignAssets []uint64,
+	appBoxReferences []types.AppBoxReference,
 	approvalProg []byte,
 	clearProg []byte,
 	sp types.SuggestedParams,
@@ -716,6 +724,7 @@ func MakeApplicationUpdateTx(
 		accounts,
 		foreignApps,
 		foreignAssets,
+		appBoxReferences,
 		types.UpdateApplicationOC,
 		approvalProg,
 		clearProg,
@@ -737,6 +746,7 @@ func MakeApplicationDeleteTx(
 	accounts []string,
 	foreignApps []uint64,
 	foreignAssets []uint64,
+	appBoxReferences []types.AppBoxReference,
 	sp types.SuggestedParams,
 	sender types.Address,
 	note []byte,
@@ -748,6 +758,7 @@ func MakeApplicationDeleteTx(
 		accounts,
 		foreignApps,
 		foreignAssets,
+		appBoxReferences,
 		types.DeleteApplicationOC,
 		nil,
 		nil,
@@ -770,6 +781,7 @@ func MakeApplicationOptInTx(
 	accounts []string,
 	foreignApps []uint64,
 	foreignAssets []uint64,
+	appBoxReferences []types.AppBoxReference,
 	sp types.SuggestedParams,
 	sender types.Address,
 	note []byte,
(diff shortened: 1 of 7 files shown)


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="feature/box-storage"
```
ahangsu, line 3: TODO: when algorand-sdk-testing PR merged:
```suggestion
SDK_TESTING_BRANCH="master"
```

test/integration.tags
```diff
@@ -1,5 +1,7 @@
 @abi
 @algod
+@applications
```
tzaffi, line 3: Is the `@applications` test necessary? There isn't presently an unqualified `@applications` tag in the SDK testing branch `feature/box-storage`

ahangsu: I am adding this tag to make the `integration.tags` uniform with others in other sdks, but if `@applications` is unnecessary, then we should just take them all out cross SDKs.

tzaffi: Ahh right. Yeah, I probly should have removed `@applications` when I refactored the repos to use sandbox. We should just make sure the same tests run before/after the tag removal.

ahangsu: the tag should gone in latest commit, and the commit should pass

ahangsu: test passing, thanks for point out unnecessary tags, I am resolving this thread here

encoding/box_query_encoding.go
```diff
@@ -0,0 +1,43 @@
...
+const encodingBase64Prefix = "b64:"
+
+// EncodeBytesForBoxQuery provides a convenience method to string encode box names for use with Box search APIs (e.g. GetApplicationBoxByName).
+func EncodeBytesForBoxQuery(xs []byte) string {
+	return encodingBase64Prefix + base64.StdEncoding.EncodeToString(xs)
+}
+
+// EncodeBoxForBoxQuery provides a convenience method to string encode box names for use with Box search APIs (e.g. GetApplicationBoxByName).
+func EncodeBoxForBoxQuery(b models.Box) (string, error) {
+	decoded, err := base64.StdEncoding.DecodeString(string(b.Name))
```
tzaffi, line 18: Is the fact that we're using `StdEncoding` (as opposed to `URLEncoding`) here and everywhere else  an artifact of the code generator? 

@michaeldiamant

future/transaction.go
```diff
@@ -985,4 +1010,43 @@ func parseTxnForeignAssets(foreignAssets []uint64) (parsed []types.AssetIndex) {
...
+					br.ForeignAppIdx = uint64(idx + 1)
+					break
+				}
+			}
+		}
+
+		if !found && abr.AppID == curAppID {
+			found = true
+			br.ForeignAppIdx = 0
+		}
```
ahangsu, line 1040: nit: want to propose a refactoring: move line 1037 check on `abr.AppID == curAppID` to line 1024, so 2 cases can essentially be merged into 1.

ahangsu: my bad, disregard this.

ahangsu (approved):
approving for parts other than `test` looks right to me

test/algodclientv2_test.go
```diff
@@ -243,6 +256,16 @@ func weMakeAnAccountApplicationInformationCallAgainstAccountApplicationID(accoun
 	return nil
 }
 
+func withClient(f func(client algod.Client)) error {
```
tzaffi, line 259:
```suggestion
func withClientAndMockServer(f func(client algod.Client)) error {
```

nit: suggest a name that clarifies to a new user of the codebase (such as me) that the request will be made against `mockServer.URL`

ahangsu: I think I would refactor this part of code, so we don't have to declare a local function under cucumber step, and the refactoring would take this `withClient` function out.

tzaffi (approved):
LGTM

client/v2/common/models/boxes_response.go
```diff
@@ -0,0 +1,10 @@
+package models
+
+// BoxesResponse box names of an application
+type BoxesResponse struct {
```
jasonpaulos, line 4: This is missing the `next-token` field. I suggest re-running the generator to pick up the latest changes.

future/transaction.go
```diff
@@ -618,6 +621,7 @@ func MakeApplicationCreateTx(
 	accounts []string,
 	foreignApps []uint64,
 	foreignAssets []uint64,
+	appBoxReferences []types.AppBoxReference,
```
jasonpaulos, line 624: Unfortunately adding a new function parameter is a breaking change, and a pretty significant one at that. This would cause all downstream code that uses one of these common functions to have a compilation error.

In the past we've had to solve this by introducing a new function, like `MakeApplicationCreateTxWithExtraPages`, which contains the new parameter, and leaving the original function unmodified.

To avoid this annoyance going forward, we really should replace these functions with either ones that accept a single "parameters" struct, or with a builder pattern, but I expect that's a bit out of scope at this time.

encoding/box_query_encoding.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return EncodeBytesForBoxQuery(decoded), nil
+}
+
+func EncodeBoxDescriptorForBoxQuery(bd models.BoxDescriptor) (string, error) {
+	decoded, err := base64.StdEncoding.DecodeString(string(bd.Name))
```
jasonpaulos, line 27: I believe that `bd.Name` will already be base64 decoded by the JSON decoder.

This is not obvious, since the `Name` field still retains the REST API description that it's a base64 encoded string. Yet that's no longer accurate here.

test/applications_integration_test.go
```diff
@@ -757,11 +808,153 @@ func checkRandomElementResult(resultIndex int, input string) error {
...
+	for i, b := range r.Boxes {
+		actualNames[i] = b.Name
+	}
+
+	if len(expectedNames) != len(actualNames) {
+		return fmt.Errorf("expected and actual box names length do not match:  %v != %v", len(expectedNames), len(actualNames))
+	}
+
+	for _, e := range expectedNames {
+		if !bytesContains(e, actualNames) {
```
jasonpaulos, line 883: This test would be stronger if we directly compare the actual and expected name slices

test/applications_integration_test.go
```diff
@@ -757,11 +808,153 @@ func checkRandomElementResult(resultIndex int, input string) error {
...
+			expectedNames[i] = expected
+		}
+	}
+
+	if len(expectedNames) != len(actualNames) {
+		return fmt.Errorf("expected and actual box names length do not match:  %v != %v", len(expectedNames), len(actualNames))
+	}
+
+	for _, e := range expectedNames {
+		if !bytesContains(e, actualNames) {
```
jasonpaulos, line 945: Same here

jasonpaulos:
I've created child PRs #410 and #411 to address my above comments -- review would be welcome

client/v2/common/models/boxes_response.go
jasonpaulos: Fixed in #411

jasonpaulos (approved):

.test-env
michaeldiamant, line 3:
```suggestion
SDK_TESTING_BRANCH="master"
```

PR #342, opened 2022-07-07
Boxes:  Add convenience methods for encoding Box names

michaeldiamant:
Adds convenience methods to ensure proper Box name encoding when querying Boxes.

Provides Go analog to https://github.com/algorand/java-algorand-sdk/pull/346.

### DIFF
--- a/encoding/box_query_encoding.go
+++ b/encoding/box_query_encoding.go
@@ -1,0 +1,33 @@
+package encoding
+
+import (
+	"encoding/base64"
+	"github.com/algorand/go-algorand-sdk/client/v2/common/models"
+	"github.com/algorand/go-algorand-sdk/types"
+)
+
+const encodingBase64Prefix = "b64:"
+
+// EncodeBytesForBoxQuery provides a convenience method to string encode box names for use with Box search APIs (e.g. GetApplicationBoxByName).
+func EncodeBytesForBoxQuery(xs []byte) string {
+	return encodingBase64Prefix + base64.StdEncoding.EncodeToString(xs)
+}
+
+// EncodeBoxForBoxQuery provides a convenience method to string encode box names for use with Box search APIs (e.g. GetApplicationBoxByName).
+func EncodeBoxForBoxQuery(b models.Box) (string, error) {
+	decoded, err := base64.StdEncoding.DecodeString(string(b.Name))
+	if err != nil {
+		return "", err
+	}
+
+	return EncodeBytesForBoxQuery(decoded), nil
+}
+
+// EncodeBoxReferenceForBoxQuery provides a convenience method to string encode box names for use with Box search APIs (e.g. GetApplicationBoxByName).
+func EncodeBoxReferenceForBoxQuery(br types.BoxReference) string {
+	return EncodeBytesForBoxQuery(br.Name)
+}
+
+func EncodeAppBoxReferenceForBoxQuery(abr types.AppBoxReference) string {
+	return EncodeBytesForBoxQuery(abr.Name)
+}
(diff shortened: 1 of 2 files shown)


encoding/box_query_encoding.go
```diff
@@ -0,0 +1,33 @@
...
+const encodingBase64Prefix = "b64:"
+
+// EncodeBytesForBoxQuery provides a convenience method to string encode box names for use with Box search APIs (e.g. GetApplicationBoxByName).
+func EncodeBytesForBoxQuery(xs []byte) string {
+	return encodingBase64Prefix + base64.StdEncoding.EncodeToString(xs)
+}
+
+// EncodeBoxForBoxQuery provides a convenience method to string encode box names for use with Box search APIs (e.g. GetApplicationBoxByName).
+func EncodeBoxForBoxQuery(b models.Box) (string, error) {
+	decoded, err := base64.StdEncoding.DecodeString(string(b.Name))
```
algoidurovic, line 18: If `models.Box.Name` is meant to be used as a b64 string then why is the type bytes? This PR doesn't need to be concerned with this though, I'm just curious.

encoding/box_query_encoding.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return EncodeBytesForBoxQuery(decoded), nil
+}
+
+// EncodeBoxReferenceForBoxQuery provides a convenience method to string encode box names for use with Box search APIs (e.g. GetApplicationBoxByName).
```
algoidurovic, line 26: I wonder if the two methods below would be better as methods belonging to their respective types.

algoidurovic (commented):
Looks good overall, just have a few questions.

encoding/box_query_encoding.go
michaeldiamant: @algoidurovic It's a good question - let me know if the response addresses it?

* The data type is bytes to match the OAS2 specification for type = _string_ + format = _bytes_  (https://swagger.io/specification/v2/ - see _Data Types_).
* It's definitely confusing.  Your question led me to closely (re-)review go-algorand.  And I briefly paired with @algochoi (thanks!) to clarify the types.

encoding/box_query_encoding_test.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type example struct {
+	source           []byte
+	expectedEncoding string
+}
+
+var e = example{[]byte("tkÿÿ"), "b64:dGvDv8O/"}
+
+func TestEncode(t *testing.T) {
```
algochoi, line 18: Is this test run in the CI? I briefly checked the logs and it seems like only the cucumber tests are run. I'm not too familiar with the repo, but I'm guessing this could also be the case with all the other `*_test.go` files

michaeldiamant: @algochoi To close the loop from our verbal - It's a bit _non_-obvious, but the non-Cucumber tests _are_ run by CI.

* The build runs `make unit` (https://github.com/algorand/go-algorand-sdk/blob/develop/Makefile#L20), which includes the non-Cucumber tests.
* Searching the full CI build output (https://circleci.com/api/v1.1/project/github/algorand/go-algorand-sdk/985/output/108/0?file=true&allocation-id=62c8753c4f212d3f212b6fba-0-build%2FDRTQJFDE) for `go test ` shows the tests running.
  * Unclear if by-design, but `make integration` also runs the non-Cucumber tests.  Explains why the output repeats.

algochoi: appreciate the clarification!

algochoi (approved):

PR #352, opened 2022-08-09
AVM: Use avm-abi repo

jasonpaulos:
PR to use the new `avm-abi` repo.

Where possible, I've removed code from this SDK that is also implemented in `avm-abi`.


abi/abi.go
```diff
@@ -1,33 +1,66 @@
 package abi
 
 import (
-	"fmt"
-	"math"
-
-	"github.com/algorand/go-algorand/data/abi"
+	avm_abi "github.com/algorand/avm-abi/abi"
```
michaeldiamant, line 4: @jasonpaulos For my understanding:  Is there a reason to prefer `avm_abi` here while https://github.com/algorand/go-algorand/pull/4375 uses `abi`?

michaeldiamant: Disregard - There's a different package structure here than go-algorand.  Resolving.

michaeldiamant (approved):

ahangsu (approved):
good to remove all these unnecessary dependencies!

PR #366, opened 2022-08-19
Enhancement: Deprecating use of langspec

ahangsu:


logic/logic.go
```diff
@@ -29,16 +32,31 @@ type operation struct {
 	Group         []string
 }
 
+// Deprecated
```
michaeldiamant, line 35: @ahangsu Minor:  Based on my read of https://github.com/golang/go/wiki/Deprecated, we ought to tweak deprecation format here.

> To signal that an identifier should not be used, add a paragraph to its doc comment that begins with `Deprecated:` followed by some information about the deprecation, and a recommendation on what to use instead, if applicable.

ahangsu: I see, that makes sense, we can extend the deprecation note from java one to the golang `// Deprecated: ...`.

.test-env
```diff
@@ -1,6 +1,7 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+# TODO revert back to "master" after sdk-testing is merged
+SDK_TESTING_BRANCH="deprecate-langspec"
```
ahangsu, line 4: after algorand-sdk-testing is merged, do:
```suggestion
SDK_TESTING_BRANCH="master"
```

logic/logic.go
```diff
@@ -29,16 +32,58 @@ type operation struct {
...
+		if !isAsciiPrintableByte(b) {
+			return false
+		}
+	}
+	return true
+}
+
+// SanityCheckProgram performs heuristic program validation:
+// check if passed in bytes are Algorand address or is B64 encoded, rather than Teal bytes
+func SanityCheckProgram(program []byte) error {
```
michaeldiamant, line 67: @ahangsu Let me know what you think of the following proposal?

I propose marking `SanityCheckProgram` (and analogous methods in other SDKs) private.

Since the check is _only_ a set of heuristics, my thinking is that we should _not_ encourage API users to rely on the check.

logic/logic.go
```diff
@@ -29,16 +32,58 @@ type operation struct {
...
+		}
+		return fmt.Errorf("program bytes are all ASCII printable characters, not looking like Teal byte code")
+	}
+	return nil
+}
+
 // ReadProgram is used to validate a program as well as extract found variables
+// Deprecated: `langspec.json` can no longer correctly to depicting the cost model (as of 2022.08.22),
+// also to minimize the work in updating SDKs per AVM release, we are deprecating`langspec.json` across all SDKs.
+// The behavior of `ReadProgram` relies on `langspec.json`. Thus, this method is being deprecated.
```
michaeldiamant, line 86: @ahangsu I appreciate the added context here.  I propose these tweaks:
* Avoid using a date - it's hard to correlate a date with a released version.  Either prefer version where deprecation begins (e.g. v1.20.0 for this repo as of writing) or leaving without specification.
* Focus message on langspec.json inaccuracies rather than maintainenance concerns.

To minimize multiple iterations, I also suggest aligning all reviewers on phrasing.  I'm happy to speak live if it'll help us reach a conclusion faster.

```suggestion
// Deprecated: Validation relies on metadata (`langspec.json`) that does not accurately represent opcode behavior across program versions.
// The behavior of `ReadProgram` relies on `langspec.json`. Thus, this method is being deprecated.
```

logic/logic.go
ahangsu: good point, I made it private in js-algorand-sdk, but LogicsigSignature in `crypto` package cannot get access to this heuristic method. Is there a preferred way to make this private, while still can be accessed in other package?

ahangsu: I made the Js/Py-sdk's `sanityCheckProgram` private or protected, but in Java/Go, it does not seem straightforward to both hide this from user and export it internally

michaeldiamant: @ahangsu Here's my thinking:

For Java, provided we agree to mark `sanityCheckProgram` private, then I suggest:
* Move definition of `sanityCheckProgram` into `LogicsigSignature`.
* Mark `public class Logic` as deprecated because it'll contain no non-deprecated functionality. 
  * Implies we can remove deprecations on specific fields + methods because the class is deprecated.

For Go, I offer the same suggestion + caveats with `SanityCheckProgram` being moved into `crypto.go`.

ahangsu: that sounds reasonable to me, working on it now.

jasonpaulos: I've only done a cursory review so far, but I'll leave my thoughts here.

It seems like we cannot avoid changing the behavior of the SDKs in some way. Indeed, that's the goal of this PR. That being the case, I'd opt to change the existing `CheckProgram` function to become `SanityCheckProgram`. That way, we do not export any new functions and we can keep the implementation in this package.

ahangsu: for function exporting issue, if we move heuristic sanity check function into `LogicsigSignature`, and make them private methods, then we are just simply adding an internal method, rather than something exported. We are deprecating `CheckProgram` anyways, I think...

jasonpaulos: Ok, I will not push further. The latest changes make sense

crypto/crypto.go
```diff
@@ -611,11 +644,7 @@ func AddressFromProgram(program []byte) types.Address {
 // 2. If no ma provides, it returns Sig delegated LogicSig
 // 3. If both sk and ma specified the function returns Multisig delegated LogicSig
 func MakeLogicSig(program []byte, args [][]byte, sk ed25519.PrivateKey, ma MultisigAccount) (lsig types.LogicSig, err error) {
```
jasonpaulos, line 646: There is an unfortunate problem. This function is deprecated and has been replaced by `MakeLogicSigAccountEscrow`, `MakeLogicSigAccountDelegated`, and `MakeLogicSigAccountDelegatedMsig`.

The good news is that `MakeLogicSigAccountDelegated` and `MakeLogicSigAccountDelegatedMsig` call into `MakeLogicSig`, so this change works for them. The bad news is that `MakeLogicSigAccountEscrow` does not call into `MakeLogicSig`, and even if we modified it to do so, it does not currently return an error response.

ahangsu: ugh, my IDE does not show that it has been deprecated, I should probably mark it deprecated as golang syntax.

I think I should be a nice dude to add the check in `MakeLogicSigAccountEscrow `, so the basic check should be reflected...

ahangsu: never mind, I just realized `MakeLogicSigAccountEscrow` returns only `LogicSigAccount`, *without error*. That is unfortunate.

jasonpaulos: Right, at the moment I think the only solution is to deprecate `MakeLogicSigAccountEscrow` and create a new function that will return an error too.

ahangsu: deprecation change reflected in [f8dab81](https://github.com/algorand/go-algorand-sdk/pull/366/commits/f8dab818abe3564d6a577a18f6ea926357cebf62), let me know if you are good with new method name

crypto/account.go
```diff
@@ -198,6 +200,20 @@ func MakeLogicSigAccountEscrow(program []byte, args [][]byte) LogicSigAccount {
 	}
 }
 
+// MakeLogicSigAccountEscrowChecked creates a new escrow LogicSigAccount.
+// The address of this account will be a hash of its program.
+func MakeLogicSigAccountEscrowChecked(program []byte, args [][]byte) (LogicSigAccount, error) {
+	if err := sanityCheckProgram(program); err != nil {
```
jasonpaulos, line 206: nit: it would be a little nicer to call `MakeLogicSig` here to unify all the code paths.

crypto/crypto.go
jasonpaulos: Thanks, new name looks good

crypto/account.go
ahangsu: yes, that makes sense

test/steps_test.go
```diff
@@ -2634,3 +2639,33 @@ func theResultingSourceMapIsTheSameAsTheJson(expectedJsonPath string) error {
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func heuristicCheckOverBytes() error {
+	_, sanityCheckError = crypto.MakeLogicSigAccountEscrowChecked(seeminglyProgram, nil)
+	if sanityCheckError != nil {
+		fmt.Printf(sanityCheckError.Error())
```
jasonpaulos, line 2655: nit: is this printf meant to be here?

jasonpaulos (commented):
This PR looks good to me, with the exception of one comment below.

I will come back and approve once the algorand-sdk-testing branch has been updated and that comment is resolved.

test/steps_test.go
ahangsu: ugh my bad, fat finger

michaeldiamant (approved):

jasonpaulos (approved):

PR #370, opened 2022-08-23
BugFix: Src map type assert fix

barnjamin:
It turns out reading json from disk and unmarshalling it produces a float whereas the result from the API returns a uint64.

```go

func main() {

	b, err := os.ReadFile("../sourcemap.json")
	if err != nil {
		log.Fatalf("%+v", err)
	}

	m := map[string]interface{}{}
	err = json.Unmarshal(b, &m)
	if err != nil {
		log.Fatalf("fudge: %+v", err)
	}

        // Works
	mp, err := logic.DecodeSourceMap(m)
	if err != nil {
		log.Fatalf("fudge %+v", err)
	}
	log.Printf("%+v", mp)

	client, err := algod.MakeClient("http://localhost:4001", strings.Repeat("a", 64))
	if err != nil {
		log.Fatalf("fudge: %+v", err)
	}

	var program []byte = []byte("#pragma version 6\npushint 1\npop\nint 1\nreturn")

	result, err := client.TealCompile(program).Sourcemap(true).Do(context.Background())
	if err != nil {
		log.Fatalf("Failed to compile: %+v", err)
	}

        // Doesn't work (prior to change)
	mp, err = logic.DecodeSourceMap(*result.Sourcemap)
	if err != nil {
		log.Fatalf("fudge %+v", err)
	}

	log.Printf("%+v", mp)

}

```


logic/source_map.go
```diff
@@ -22,7 +22,12 @@ func DecodeSourceMap(ism map[string]interface{}) (SourceMap, error) {
 	sm := SourceMap{}
```
winder, line 22: Alternative:
```
	buf:=json.Encode(ism)
	json.Decode(buf,&sm)
```

barnjamin:
@winder 
> Why are you unmarshaling to a map[string]interface{} instead of a structured object?

You mean why does this method accept a `map[string]interface{}` ? If so, its just what we get from the call to compile endpoint: https://github.com/algorand/go-algorand-sdk/blob/develop/client/v2/common/models/compile_response.go#L12

logic/source_map.go
barnjamin: Can you help me understand why this is better?

winder: The only reason I bring it up, is that it's easy to get things wrong with reflection. So if there's a way to avoid using it the code would be easier to reason about and less likely that there is an error. So basically, I'm asking if these are simpler alternatives.

> Why are you unmarshaling to a map[string]interface{} instead of a structured object?

What I mean is you could use Unmarshal into a real type instead of `map[string]interface{}`, that way `Unmarshal` would handle all of this logic. Looking a bit closer just now, I wonder if this is even better:
```
func DecodeSourceMap(ism map[string]interface{}) (SourceMap, error) {
  buf := json.Encode(ism)
  var sourcemap SourceMap
  return json.Decode(buf, sourcemap)
}
```

barnjamin: I can do that to begin with to set the fields specified but I still need to do some post-processing to decode the VLQ values. Is it worth it?

winder: I think it is, but not enough to block your PR so I'll leave it up to you.

winder (approved):

logic/source_map.go
barnjamin: I'm going to merge this and post a follow up with the above since I think this is tripping folks up right now

PR #362, opened 2022-08-17
StateProof: State proof support

winder:
Add StateProof support to go SDK.

PR #374, opened 2022-08-25
StateProof: State proof support

Eric-Warehime:
Picks up where #362 leaves off since Will is out.

I've addressed all of the most recent comments in that PR--added allocbound definitions where missing, renamed fields.


client/v2/algod/algod.go
```diff
@@ -80,8 +80,8 @@ func (c *Client) Block(round uint64) *Block {
 	return &Block{c: c, round: round}
 }
 
-func (c *Client) GetProof(round uint64, txid string) *GetProof {
-	return &GetProof{c: c, round: round, txid: txid}
+func (c *Client) GetTransactionProof(round uint64, txid string) *GetTransactionProof {
```
algoidan, line 83: I strongly suggest that we should consider adding the hash type as a parameter even though it is an optional parameter to the REST.
Bridge operators would most use the SHA256 proofs and not the default one.

client/v2/common/models/transaction_proof_response.go
```diff
@@ -0,0 +1,22 @@
...
+	// Proof proof of transaction membership.
+	Proof []byte `json:"proof"`
+
+	// Stibhash hash of SignedTxnInBlock for verifying proof.
+	Stibhash []byte `json:"stibhash"`
+
+	// Treedepth represents the depth of the tree that is being proven, i.e. the number
+	// of edges from a leaf to the root.
+	Treedepth uint64 `json:"treedepth"`
+}
```
algoidan, line 22: is it true that we do not need `client/v2/common/models/proof_response.go`? I think this covers for it

types/block.go
```diff
@@ -93,11 +89,28 @@ type (
...
 		ParticipationUpdates
 	}
 
+	// TxnCommitments represents the commitments computed from the transactions in the block.
+	// It contains multiple commitments based on different algorithms and hash functions, to support different use cases.
+	TxnCommitments struct {
+		_struct struct{} `codec:",omitempty,omitemptyarray"`
+		// Root of transaction merkle tree using SHA512_256 hash function.
+		// This commitment is computed based on the PaysetCommit type specified in the block's consensus protocol.
+		NativeSha512_256Commitment GenericDigest `codec:"txn"`
```
algoidan, line 108: those two fields are defined as `Digest` on go-algorand. I think it should be defined in the same way.

types/block.go
```diff
@@ -93,11 +89,28 @@ type (
 		// started being supported).
 		TxnCounter uint64 `codec:"tc"`
 
+		// StateProofTracking tracks the status of the state proofs, potentially
+		// for multiple types of ASPs (Algorand's State Proofs).
+		//msgp:sort protocol.StateProofType protocol.SortStateProofType
+		StateProofTracking map[StateProofType]StateProofTrackingData `codec:"spt,allocbound=protocol.NumStateProofTypes"`
```
algoidan, line 95: I think `NumStateProofTypes`  is being referenced from the wrong package.

types/stateproof.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// Message represents the message that the state proofs are attesting to. This message can be
+// used by lightweight client and gives it the ability to verify proofs on the Algorand's state.
+// In addition to that proof, this message also contains fields that
+// are needed in order to verify the next state proofs (VotersCommitment and LnProvenWeight).
+type Message struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+	// BlockHeadersCommitment contains a commitment on all light block headers within a state proof interval.
+	BlockHeadersCommitment []byte `codec:"b,allocbound=crypto.Sha256Size"`
```
algoidan, line 214: I think `Sha256Size` is not defined under crypto package. same below

algoidan (commented):
Overall looks very good. Thanks!
I've added some comment to take into account

client/v2/algod/algod.go
algoidan, line 83: is there a way to easily invoke this function with the non-default hashtype param?

Eric-Warehime: `GetTransactionProof(round, txid).HashType("sha512_256").Do(args)`

Eric-Warehime: See https://github.com/algorand/go-algorand-sdk/pull/374/files/0842b7ff7bbfc39426b42411b65a8cc707404b33#diff-54d7a8bea6031095db3bfcc1351bf79d7289963f51ba21f268b66684b41bd142R36

algoidan (approved):

PR #377, opened 2022-08-28
StateProof: State Proof Verification additions

almog-t:


go.mod
```diff
@@ -4,18 +4,21 @@ go 1.17
 
 require (
 	github.com/algorand/avm-abi v0.1.0
-	github.com/algorand/go-codec/codec v1.1.8
+	github.com/algorand/go-codec/codec v1.1.9
```
algoidan, line 7: not sure we need to upgrade this at the moment (also for testify)

stateproofs/stateProofVerifier.go
```diff
@@ -0,0 +1,32 @@
...
+type StateProofVerifier struct {
+	stateProofVerifier *stateproof.Verifier
+}
+
+func InitializeVerifier(votersCommitment types.GenericDigest, lnProvenWeight uint64) *StateProofVerifier {
+	return &StateProofVerifier{stateProofVerifier: stateproof.MkVerifierWithLnProvenWeight(stateprooftypes.GenericDigest(votersCommitment),
+		lnProvenWeight, strengthTarget)}
+}
+
+func (v *StateProofVerifier) VerifyStateProofMessage(stateProof *types.EncodedStateProof, message *types.Message) error {
```
algoidan, line 22: Think `VerifyStateProof` or `Verify` would be a better name

types/lightBlockHeader.go
```diff
@@ -0,0 +1,28 @@
...
+type LightBlockHeader struct {
+	_struct struct{} `codec:",omitempty,omitemptyarray"`
+
+	Seed                Seed   `codec:"0"`
+	RoundNumber         Round  `codec:"r"`
+	GenesisHash         Digest `codec:"gh"`
+	Sha256TxnCommitment Digest `codec:"tc,allocbound=Sha256Size"`
+}
+
+// ToBeHashed implements the crypto.Hashable interface
```
algoidan, line 25: Think we can move this into crypto package to be aligned with sdk ordering of the packages

types/hashes.go
```diff
@@ -0,0 +1,37 @@
...
+type GenericDigest []byte
+
+// HashID is a domain separation prefix for an object type that might be hashed
+// This ensures, for example, the hash of a transaction will never collide with the hash of a vote
+type HashID string
+
+// Hashable is an interface implemented by an object that can be represented
+// with a sequence of bytes to be hashed or signed, together with a type ID
+// to distinguish different types of objects.
+type Hashable interface {
```
algoidan, line 16: This generic type is very useful on go-algorand, but I think we can use a simple version here. i.e let's just create a function that hashes the stateproof message and the light block header.

stateproofs/stateProofVerifier_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	err = json.Decode(contents, &target)
+	assertions.NoError(err)
+}
+
+func TestStateProofVerification(t *testing.T) {
+	a := require.New(t)
+
+	prevStateProofFileName := "prevStateProof.json"
+	newStateProofFileName := "newStateProof.json"
```
winder, line 26: You can use `embed` to load these files

types/stateproof.go
```diff
@@ -1,8 +1,22 @@
...
+type MessageHash [32]byte
+
 // StateProofType identifies a particular configuration of state proofs.
 type StateProofType uint64
 
+const MaxHashDigestSize = sumhash.Sumhash512DigestSize
+
+// GenericDigest is a digest that implements CustomSizeDigest, and can be used as hash output.
+//msgp:allocbound GenericDigest MaxHashDigestSize
+type GenericDigest []byte
```
winder, line 18: Since you're adding this back in, could you update the block type to use it as well?

stateproofs/stateProofVerifier.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/algorand/go-stateproof-verification/stateproofcrypto"
+
+	"github.com/algorand/go-algorand-sdk/crypto"
+	"github.com/algorand/go-algorand-sdk/encoding/msgpack"
+	"github.com/algorand/go-algorand-sdk/types"
+)
+
+const strengthTarget = uint64(256)
+
+type StateProofVerifier struct {
```
winder, line 14: I think you should move this code into the `go-stateproof-verification` package.

You can avoid exporting types from that package by aliases the library type in this package:
```
type StateProofVerifier = stateproof.StateProofVerifier
```

almog-t: The rational for including this code in the SDK is allowing users to stick with the SDK (directly at least), rather than adding another library.

algoidan (approved):
LGTM

types/stateproof.go
winder: I specifically removed this type at the last minute after Idan suggested I avoid including too many SP specific types. If it's being added back you can revert this change:
https://github.com/algorand/go-algorand-sdk/pull/378/files#diff-77409c5f41482f194140901cdad16d5bf6e5e6b3591f275dbf3660f12615078cL193

stateproofs/stateProofVerifier.go
```diff
@@ -0,0 +1,33 @@
+package stateproofverification
```
winder, line 1: It seems weird that the package name is not the same as the directory name, could you rename the package or the directory?

stateproofs/stateProofVerifier_test.go
almog-t: Perhaps I'm missing something, but using embed causes CI to fail for go 1.15. It seems I can't do that.

winder: I don't think you're missing anything, let's remove 1.15 from CI.

almog-t: Done.

winder (approved):

PR #378, opened 2022-08-29
State Proofs: Use generic type for StateProof txn field.

winder:
These types would be available in the forthcoming state proof verification library.


algoidan (approved):
Looks great
thanks

PR #380, opened 2022-08-30
Improvement: Better SourceMap decoding

barnjamin:
From wills suggestion here: https://github.com/algorand/go-algorand-sdk/pull/370#discussion_r958809876


winder (approved):
Thank you!

PR #392, opened 2022-09-07
Add deprecation tags to v1 API

algochoi:
This PR adds deprecated tags to the v1 `algod` client. 

After "sunsetting" the deprecated code, any unit tests and cucumber steps that use them should be deleted as well.

### DIFF
--- a/client/v2/common/models/dryrun_txn_result.go
+++ b/client/v2/common/models/dryrun_txn_result.go
@@ -15,7 +15,8 @@ type DryrunTxnResult struct {
 	// BudgetConsumed budget consumed during execution of app call transaction.
 	BudgetConsumed uint64 `json:"budget-consumed,omitempty"`
 
-	// Cost net cost of app execution. Field is DEPRECATED and is subject for removal.
+	// Cost net cost of app execution.
+	// Deprecated: this field is deprecated and is subject for removal.
 	// Instead, use `budget-added` and `budget-consumed.
 	Cost uint64 `json:"cost,omitempty"`
 
--- a/client/algod/algod.go
+++ b/client/algod/algod.go
@@ -1,5 +1,7 @@
 package algod
 
+// Deprecated: v1 algod client is deprecated, please use the v2 algod client.
+
 import (
 	"bytes"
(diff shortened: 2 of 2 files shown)


client/v2/common/models/dryrun_txn_result.go
```diff
@@ -15,7 +15,8 @@ type DryrunTxnResult struct {
 	// BudgetConsumed budget consumed during execution of app call transaction.
 	BudgetConsumed uint64 `json:"budget-consumed,omitempty"`
 
-	// Cost net cost of app execution. Field is DEPRECATED and is subject for removal.
+	// Cost net cost of app execution.
+	// Deprecated: this field is deprecated and is subject for removal.
```
michaeldiamant, line 19: @algochoi Does the highlighted change remain unchanged after running _generator_?

* My local testing suggests _no_.
* I think the PR should make no change that conflicts with _generator_.  If you agree, then I think the change should be reverted.
* The intent is good - The PR attempts to follow Go's convention for applying deprecations.  I think a root resolution is to modify _generator_ to support language-specific deprecations.
  * My opinion is that it's _not_ worth trying to address the concern in-scope to the PR.  I suspect it's a non-trivial lift.
  * Optionally, if you've agreed with the thinking, I suggest opening an _issue_ in https://github.com/algorand/generator to highlight the gap.

algochoi: Good point, thank you for catching - I agree with your thoughts and reverted the change in the latest commit

algochoi: Documented it here: https://github.com/algorand/generator/issues/41

michaeldiamant: @algochoi Resolving - Thanks for opening the issue as future reference point + the revert.

michaeldiamant (approved):

ahangsu (commented):
minor: I would suggest mark `client/wrappers.go` also deprecated, for most of the methods from `client Client` relies on v1 API

ahangsu (approved):

PR #400, opened 2022-09-14
BugFix: Fix dryrun parser

barnjamin:
Duplicate work, also previously was passing the wrong var into json.Marshal

### DIFF
--- a/future/dryrun.go
+++ b/future/dryrun.go
@@ -154,6 +154,7 @@ type DryrunResponse struct {
 }
 
 func NewDryrunResponse(d models.DryrunResponse) (DryrunResponse, error) {
+	// Marshal and unmarshal to fix integer types.
 	b, err := json.Marshal(d)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


future/dryrun.go
```diff
@@ -154,14 +154,11 @@ type DryrunResponse struct {
 }
 
 func NewDryrunResponse(d models.DryrunResponse) (DryrunResponse, error) {
-	// TODO:  this is lazy but also works?
-	dr := DryrunResponse{}
-	b, err := json.Marshal(dr)
+	b, err := json.Marshal(d)
```
winder, line 157: Is this right? We should have a comment that explains why we're doing this thing.
```suggestion
	// Marshal and unmarshal to fix integer types.
	b, err := json.Marshal(d)
```

Eric-Warehime (approved):
I don't have a ton of context on the integer type problem that's being solved by this function, but the new code is better than the old code so LGTM in that sense.

PR #408, opened 2022-09-21
Minor refactoring in code implementation

ahangsu:


future/transaction.go
```diff
@@ -1021,7 +1022,7 @@ func parseBoxReferences(abrs []types.AppBoxReference, foreignApps []uint64, curA
 		br := types.BoxReference{Name: abr.Name}
 		found := false
 
-		if abr.AppID == 0 {
+		if abr.AppID == 0 || abr.AppID == curAppID {
```
tzaffi, line 1025: This change is now out of sync with part (2) of the comment. In particular, when `abr.AppID == curAppID`, then `ForeignAppIdx` is set to 0 regardless of whether it's provided in the foreign apps array.

I don't know enough about the functionality to have an opinion on which approach is preferred, but the comment should be modified if the proposed change is kept.

ahangsu: I overlooked the code test cases and mistaken on how the code works. Disregard this PR and my apologies for the mistake.

Issue #397, opened 2022-09-12
Breaking changes in v3.0.0


michaeldiamant:
Itemizes functionality that will be removed in the upcoming major release.

---

Migrating a few thoughts from v2.0.0.  If we'd like to pursue, we can add stories.

* Modify `transaction` construction to use builders (or similar) in order to minimize backwards incompatible changes. 
  * Related:  https://algorand.atlassian.net/wiki/spaces/FDE/pages/491716666/SuggestedParams+Thoughts#Go 
* Remove special cases from HTTP Clients:
  * Update generator templates to remove special cases.
  * Regenerate client libraries.
* Split algod and indexer generated types: https://github.com/algorand/generator/issues/59

PR #410, opened 2022-09-23
Box storage improvements from review

jasonpaulos:
Child PR to #341

This PR would resolve all of my review comments, except for https://github.com/algorand/go-algorand-sdk/pull/341#discussion_r979061205, which is addressed by #411

### DIFF
--- a/future/transaction.go
+++ b/future/transaction.go
@@ -612,7 +612,7 @@ func byte64FromBase64(in string) (out [64]byte, err error) {
 // MakeApplicationCreateTx makes a transaction for creating an application (see above for args desc.)
 // - optIn: true for opting in on complete, false for no-op.
 //
-// NOTE: if you need to use extra pages or boxes, use MakeApplicationUpdateTxWithBoxes instead.
+// NOTE: if you need to use extra pages or boxes, use MakeApplicationCreateTxWithBoxes instead.
 func MakeApplicationCreateTx(
 	optIn bool,
 	approvalProg []byte,
@@ -653,7 +653,7 @@ func MakeApplicationCreateTx(
 // MakeApplicationCreateTxWithExtraPages makes a transaction for creating an application (see above for args desc.)
 // - optIn: true for opting in on complete, false for no-op.
 //
-// NOTE: if you need to use boxes, use MakeApplicationUpdateTxWithBoxes instead.
+// NOTE: if you need to use boxes, use MakeApplicationCreateTxWithBoxes instead.
 func MakeApplicationCreateTxWithExtraPages(
 	optIn bool,
 	approvalProg []byte,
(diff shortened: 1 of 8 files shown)


future/transaction.go
```diff
@@ -621,31 +623,66 @@ func MakeApplicationCreateTx(
...
+		group,
+		lease,
+		rekeyTo,
+	)
+}
+
+// MakeApplicationCreateTxWithExtraPages makes a transaction for creating an application (see above for args desc.)
+// - optIn: true for opting in on complete, false for no-op.
+//
+// NOTE: if you need to use boxes, use MakeApplicationUpdateTxWithBoxes instead.
```
michaeldiamant, line 656: @jasonpaulos Minor:  Seems the comment's intent is to reference _create_ instead of _update_. Can you double check me?

```suggestion
// NOTE: if you need to use boxes, use MakeApplicationCreateTxWithBoxes instead.
```

ahangsu (approved):
The change looks right, thanks for noticing the breaking change in transaction creation methods, now I really wish to have some default argument in golang...

future/transaction.go
```diff
@@ -597,20 +597,22 @@ func byte64FromBase64(in string) (out [64]byte, err error) {
...
-//                 A page is 1024 bytes. This field enables execution of app programs
-//                 larger than the default maximum program size.
-//
 // - boxes         lists the boxes to be accessed during evaluation of the application
 //                 call. This also must include the boxes accessed by inner app calls.
 
 // MakeApplicationCreateTx makes a transaction for creating an application (see above for args desc.)
 // - optIn: true for opting in on complete, false for no-op.
+//
+// NOTE: if you need to use extra pages or boxes, use MakeApplicationUpdateTxWithBoxes instead.
```
ahangsu, line 615: I think @michaeldiamant 's comment also apply here tho...

future/transaction.go
ahangsu: I think you are right, there is one other case need to be taken care of, and the rest is fine.

jasonpaulos: Thanks for noticing, this was a copy-paste error. It's fixed in 9b8ad96a29b8b01fe3ddf129e0249f4f2b7f6e18

future/transaction.go
jasonpaulos: Thanks, fixed in 9b8ad96a29b8b01fe3ddf129e0249f4f2b7f6e18

michaeldiamant (approved):
@jasonpaulos Thanks for catching the backwards incompatibility + base64 correctness concern.

Hopefully in v3.0.0 (https://github.com/algorand/go-algorand-sdk/issues/397), we'll make a pass at improving Transaction construction.

PR #427, opened 2022-10-30
codec checksum mismatch

guoshijiang:


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.17
 
 require (
 	github.com/algorand/avm-abi v0.1.0
-	github.com/algorand/go-codec/codec v1.1.8
+	github.com/algorand/go-codec/codec v1.1.9
```
michaeldiamant, line 7: @guoshijiang Thanks for opening a PR.  For my understanding, can you summarize what motivates the change either here or in the PR description?

I'm trying to gauge if you ran into an issue that required upgrading codec versions or something else.  I didn't follow what _mismatch_ refers to in the PR title.

guoshijiang: Ok, I check it

PR #428, opened 2022-10-31
Regenerate sources from upstream AVM branches

michaeldiamant:
Regenerates sources from upstream go-algorand + indexer AVM branches and brings in `develop` for the final time prior to merging the parent branch.


michaeldiamant:
Closing in favor of https://github.com/algorand/go-algorand-sdk/pull/429.

PR #434, opened 2022-11-09
Tests: Migrate v1 algod dependencies to v2 in cucumber tests

algochoi:
Migrates v1 algod usage in cucumber integration tests to use v2. Also deletes some unused steps relating to v1 APIs.

SDK Testing PR: https://github.com/algorand/algorand-sdk-testing/pull/249

### DIFF
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="remove-v1-tests"
 SDK_TESTING_HARNESS="test-harness"
 
 INSTALL_ONLY=0
--- a/.test-env
+++ b/.test-env
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="remove-v1-tests"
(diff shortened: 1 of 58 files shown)


.test-env
```diff
@@ -1,6 +1,6 @@
 # Configs for testing repo download:
 SDK_TESTING_URL="https://github.com/algorand/algorand-sdk-testing"
-SDK_TESTING_BRANCH="master"
+SDK_TESTING_BRANCH="remove-v1-tests"
```
algochoi, line 3:
```suggestion
SDK_TESTING_BRANCH="master"
```

test/steps_test.go
```diff
@@ -1555,37 +1422,54 @@ func checkExpectedVsActualAssetParams() error {
...
+		if idx == assetID {
+			found = true
+			foundAsset = asset
+			break
+		}
+	}
+	if !found {
+		return modelsV2.AssetHolding{}, fmt.Errorf("asset ID %d was not found", assetID)
+	}
+	return foundAsset, nil
```
ahangsu, line 1461: nit: simplification, 

```suggestion
	for _, asset := range assets {
		if asset.AssetId == assetID {
			return asset, nil
		}
	}
	return modelsV2.AssetHolding{}, fmt.Errorf("asset ID %d was not found", assetID)
```

ahangsu (approved):
The change seems straightforward, looks generally good to me

test/steps_test.go
algochoi: Nice catch, thanks!

PR #436, opened 2022-11-18
BugFix: Fix disassemble endpoint

zyablitsev:
POST should be used instead of GET on "/v2/teal/disassemble".

### DIFF
--- a/client/v2/algod/tealDisassemble.go
+++ b/client/v2/algod/tealDisassemble.go
@@ -18,6 +18,6 @@ type TealDisassemble struct {
 
 // Do performs the HTTP request
 func (s *TealDisassemble) Do(ctx context.Context, headers ...*common.Header) (response models.DisassembleResponse, err error) {
-	err = s.c.get(ctx, &response, "/v2/teal/disassemble", nil, headers)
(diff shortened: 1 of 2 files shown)


algochoi:
Thanks for pointing this out! This is a generated file, so we would have to override the Go definition in this file and re-generate the code here: https://github.com/algorand/generator/blob/master/go_templates/query.vm. I have a PR here: https://github.com/algorand/generator/pull/44

zyablitsev:
> Thanks for pointing this out! This is a generated file, so we would have to override the Go definition in this file and re-generate the code here: https://github.com/algorand/generator/blob/master/go_templates/query.vm. I have a PR here: [algorand/generator#44](https://github.com/algorand/generator/pull/44)

@algochoi thank you, i just used `algorand/generator` with your pr changes, looks exactly what is needed.

`client/v2/common/common.go` isn't generated, right? so we stil need to apply changes to `rawRequestPaths`

zyablitsev:
i've just merged `develop` into this pull request, please check it @algochoi

algochoi (approved):

algochoi:
CI is passing and checks look good - thanks for your work here!

PR #438, opened 2022-11-28
Enhancement: allowing zero length static array

ahangsu:
Fixes https://github.com/algorand/js-algorand-sdk/issues/497


michaeldiamant (approved):
@ahangsu Dependency bump looks as expected.

PR #447, opened 2022-12-13
tests: Check for non-empty expected err string

algochoi:
This PR changes a test check so that if the expected error string is empty, it fails the check if there is an actual error.

Currently, checking whether an empty error string is a substring of a non-empty actual error string will always return true, which is not intended behavior.

### DIFF
--- a/test/applications_integration_test.go
+++ b/test/applications_integration_test.go
@@ -786,7 +786,9 @@ func theContentsOfTheBoxWithNameShouldBeIfThereIsAnErrorItIs(fromClient, encoded
 		err = fmt.Errorf("expecting algod or indexer, got " + fromClient)
 	}
 	if err != nil {
-		if strings.Contains(err.Error(), errStr) {
+		// If the expected error string is not empty, check if it is a substring of the actual error string.
+		// Note that if the expected error string is empty, then the second condition will always return true.
+		if errStr != "" && strings.Contains(err.Error(), errStr) {
 			return nil
 		}
 		return err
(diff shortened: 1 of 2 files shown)


test/applications_integration_test.go
```diff
@@ -785,7 +785,7 @@ func theContentsOfTheBoxWithNameShouldBeIfThereIsAnErrorItIs(fromClient, encoded
 	} else {
 		err = fmt.Errorf("expecting algod or indexer, got " + fromClient)
 	}
-	if err != nil {
+	if err != nil && len(errStr) != 0 {
```
jasonpaulos, line 788: I believe you'll also want to error if `len(errStr) != 0` but `err` is nil, since you expected an error that wasn't there

algochoi: good point! Fixed in the latest commits

test/applications_integration_test.go
```diff
@@ -786,7 +786,9 @@ func theContentsOfTheBoxWithNameShouldBeIfThereIsAnErrorItIs(fromClient, encoded
 		err = fmt.Errorf("expecting algod or indexer, got " + fromClient)
 	}
 	if err != nil {
-		if strings.Contains(err.Error(), errStr) {
+		// If the expected error string is not empty, check if it is a substring of the actual error string.
+		// Note that if the expected error string is empty, then the second condition will always return true.
+		if len(errStr) != 0 && strings.Contains(err.Error(), errStr) {
 			return nil
 		}
 		return err
```
jasonpaulos, line 794: This improvement solves the problem at hand, but we should also fail if we expected an error and none was reported. Something like this:

```go
if err != nil {
  // existing if statement body, guaranteed to return
}
if len(errStr) != 0 {
  return errors.New("expected an error but none was reported")
}
```

algochoi: Agreed, good point - added check for this in [7d6e2a1](https://github.com/algorand/go-algorand-sdk/pull/447/commits/7d6e2a1c3f04390f7393966c284fd2af8aa3666c)

jasonpaulos (approved):

PR #449, opened 2022-12-13
v2: Breaking changes from v1 to v2.0.0

algochoi:
Changes supporting v2.0.0 release as discussed in https://github.com/algorand/go-algorand-sdk/issues/343.

PR #453, opened 2022-12-15
Add v2.0.0 CHANGELOG entry for DryrunTxnResult.Cost

michaeldiamant:
Extends #449 with a CHANGELOG entry for https://github.com/algorand/go-algorand-sdk/pull/444.


algochoi (approved):
thanks for the change!

PR #456, opened 2022-12-16
v2: Prepare module for v2 release

michaeldiamant:
Follows #449 to prepare go-algorand-sdk Go module for a v2 release by updating its module definition.  

* See https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher for context.
* Source file changes made mechanically via ` find . -type f -exec sed -i "" 's|"github.com/algorand/go-algorand-sdk/|"github.com/algorand/go-algorand-sdk/v2/|g' {} \;`.


go.mod
```diff
@@ -14,6 +14,5 @@ require (
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	golang.org/x/sys v0.0.0-20211019181941-9d821ace8654 // indirect
```
algochoi, line 17: Is this from running `go mod tidy`?

algochoi (approved):
Looks reasonable - example of bumping versions in the wild: https://github.com/uber-go/tally/blob/master/go.mod

go.mod
michaeldiamant: @algochoi Correct.

ahangsu (approved):
I assume we are taking **Major branch** option in https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher, rather than creating major subdirectory, then this PR should look good to me.

jasonpaulos (approved):
The changes match what I was expecting

PR #461, opened 2023-01-24
bugfix: Fix wrong response error type.

winder:
I noticed an endpoint return a "Not Found" message and a nil error. We were returning the wrong error here.

### DIFF
--- a/client/v2/common/common.go
+++ b/client/v2/common/common.go
@@ -181,7 +181,7 @@ func (client *Client) submitForm(ctx context.Context, response interface{}, path
 	// The caller wants a string
 	if strResponse, ok := response.(*string); ok {
 		*strResponse = string(bodyBytes)
-		return err
+		return responseErr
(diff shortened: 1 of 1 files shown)


client/v2/common/common.go
```diff
@@ -181,7 +181,7 @@ func (client *Client) submitForm(ctx context.Context, response interface{}, path
 	// The caller wants a string
 	if strResponse, ok := response.(*string); ok {
 		*strResponse = string(bodyBytes)
-		return err
+		return responseErr
```
algochoi, line 184: Is the `strResponse` above this line used?

algochoi: Gotcha, appreciate the example program too 👍

algochoi (approved):
Looks like a pretty subtle bug - I wonder if there are linters to warn this kinds of stuff 😅

PR #463, opened 2023-01-31
enhancement: address and blockhash encoding

shiqizng:
This PR implements marshal/unmarshal interface for Address and Blockhash data types so bookkeeping.Block can be deserialized to sdk.Block. This change is need for Indexer.  It gets block of type bookkeeping.Block from algod but it needs sdk.Block type.

### DIFF
--- a/types/address.go
+++ b/types/address.go
@@ -34,10 +34,10 @@ func (a Address) IsZero() bool {
 	return a == ZeroAddress
 }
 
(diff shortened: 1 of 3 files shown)


bbroder-algo:
Test coverage in basics_test and address_test and a new file for blockhash_test would be good to have in this enhancement for indexer

winder:
closed in favor of #464

PR #467, opened 2023-02-09
API: Add types for ledgercore.StateDelta.

winder:
* Add types for LedgerStateDelta.
* Add generated types for new status fields.
* Add strict encoder, currently only used by tests.
* Fix bug where encoding errors were ignored.
* Disable block header response test which has never worked, opened issue: https://github.com/algorand/go-algorand-sdk/issues/469
* Map keys are encoded in an unexpected way leading to a type mismatch in the test code, opened issue: https://github.com/ugorji/go/issues/390
* Disable `@unit.stateproof.responses.msgp` due to the above two issues.
* Created a "skipped" test in `json_encode.go` as another reminder (in addition to the above issues).


shiqizng:
please  add these definitions, too

```
const (
	// TealBytesType represents the type of a byte slice in a TEAL program
	TealBytesType TealType = 1

	// TealUintType represents the type of a uint in a TEAL program
	TealUintType TealType = 2
)
```

encoding/json/json_test.go
```diff
@@ -54,7 +55,32 @@ func TestDecode(t *testing.T) {
...
+
+	t.Run("strict encode map key as string", func(t *testing.T) {
+		intMap := map[int]string{
+			0: "int key",
+		}
+		data := string(EncodeStrict(intMap))
+		assert.NotContains(t, data, "0:")
+	})
+
+	t.Run("strict encode map key as string", func(t *testing.T) {
```
winder, line 78:
```suggestion
	t.Run("strict encode map interface key as string", func(t *testing.T) {
```

client/v2/common/models/shim.go
```diff
@@ -17,3 +19,6 @@ type Supply SupplyResponse
 
 // VersionBuild
 type VersionBuild BuildVersion
+
+// LedgerStateDelta aliases to the manually defined type.
+type LedgerStateDelta types.LedgerStateDelta
```
winder, line 24: I hadn't used this trick before, but it's the reason no custom generation was needed for `getLedgerStateDelta.go`. I'm not sure why the generator didn't create a `model.LedgerStateDelta` object.

Eric-Warehime (approved):

client/v2/algod/getLedgerStateDelta.go
```diff
@@ -8,15 +8,24 @@ import (
...
 
 	round uint64
+
+	p GetLedgerStateDeltaParams
 }
 
 // Do performs the HTTP request
 func (s *GetLedgerStateDelta) Do(ctx context.Context, headers ...*common.Header) (response models.LedgerStateDelta, err error) {
-	err = s.c.get(ctx, &response, fmt.Sprintf("/v2/deltas/%s", common.EscapeParams(s.round)...), nil, headers)
+	err = s.c.get(ctx, &response, fmt.Sprintf("/v2/deltas/%s", common.EscapeParams(s.round)...), s.p, headers)
```
shiqizng, line 29: is there an endpoint to get the raw deltas?

shiqizng: I'm confused by this response type, `response models.LedgerStateDelta`. I see that struct is deleted in this PR? and isn't `/v2/deltas/` in algod returning the raw state delta?

shiqizng (approved):

Eric-Warehime (approved):

PR #491, opened 2023-03-23
Docs: Examples

barnjamin:
Tweaks, reorganize examples, add smoke test to CI


examples/smoke_test.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
winder, line 1: nit: we standardized on `#!/usr/bin/env bash`

winder (approved):
LGTM

PR #520, opened 2023-04-27
Add MakeClientWithTransport client override that allows the user to pass a custom http transport

pbennett:
The sdk previously only used the default http client and the default client, using the default transport has severe limitations for apps making multiple connections to the same host. This new method will allow users to pass a new Transport instance which has new values for things like MaxIdleConnsPerHost and MaxIdleConns.

### DIFF
--- a/client/v2/algod/algod.go
+++ b/client/v2/algod/algod.go
@@ -2,6 +2,7 @@ package algod
 
 import (
 	"context"
+	"net/http"
 
 	"github.com/algorand/go-algorand-sdk/v2/client/v2/common"
 	"github.com/algorand/go-algorand-sdk/v2/client/v2/common/models"
@@ -50,6 +51,14 @@ func MakeClient(address string, apiToken string) (c *Client, err error) {
 func MakeClientWithHeaders(address string, apiToken string, headers []*common.Header) (c *Client, err error) {
 	commonClientWithHeaders, err := common.MakeClientWithHeaders(address, authHeader, apiToken, headers)
 	c = (*Client)(commonClientWithHeaders)
+	return
+}
+
+// MakeClientWithTransport is the factory for constructing a Client for a given endpoint with a
+// custom HTTP Transport as well as optional additional user defined headers.
+func MakeClientWithTransport(address string, apiToken string, headers []*common.Header, transport http.RoundTripper) (c *Client, err error) {
+	commonClientWithTransport, err := common.MakeClientWithTransport(address, authHeader, apiToken, headers, transport)
+	c = (*Client)(commonClientWithTransport)
 	return
 }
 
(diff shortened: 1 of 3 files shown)


client/v2/common/common.go
```diff
@@ -139,7 +154,7 @@ func (client *Client) submitFormRaw(ctx context.Context, path string, params int
 
 	// Supply the client token.
 	req.Header.Set(client.apiHeader, client.apiToken)
-	// Add the client headers.
+	// Add the client headers.π
```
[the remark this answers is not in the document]
bbroder-algo: no, π

bbroder-algo (dismissed):
Looks like a good idea to me, would like to see some use of it in algodclientv2_test.go

client/v2/algod/algod.go
```diff
@@ -53,6 +54,14 @@ func MakeClientWithHeaders(address string, apiToken string, headers []*common.He
 	return
 }
 
+// MakeClientWithTransport is the factory for constructing a Client for a given endpoint with a
+// custom HTTP Transport as well as optional additional user defined headers.
+func MakeClientWithTransport(address string, apiToken string, headers []*common.Header, transport http.RoundTripper) (c *Client, err error) {
```
algochoi, line 59: I believe this file is entirely auto-generated, so we'd have to make a modification in the generator repo here: https://github.com/algorand/generator/blob/master/go_templates/client.vm

client/v2/common/common.go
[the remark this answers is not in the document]
pbennett: lol - darn hotkeys w/ focus in wrong place.  I'll fix.  😀

client/v2/algod/algod.go
pbennett: Yeah - I put it in so I could use it cleanly in code pulling this fork.
The weird go-algorand -> sdk[s] relationship I'll leave to you guys.

pbennett:
There are probably better summaries but here's one I found quickly describing why this is so important:
https://www.loginradius.com/blog/engineering/tune-the-go-http-client-for-high-performance/

For backend services using multiple goroutines which make many requests to algod instances (which will all be the same url), this should allow for a massive speed improvement.

client/v2/algod/algod.go
algochoi: Created https://github.com/TxnLab/go-algorand-sdk/pull/1 for auto-generation

winder (approved):
I definitely disagree about adding a cucumber test for this. The transport interface is specific to the go SDK.

@bbroder-algo could you elaborate about what you had in mind? At most I'd think a unit test in `common_test.go` would be sufficient.

client/v2/algod/algod.go
pbennett: sorry - I missed this.  I'll merge.

algochoi (approved):

PR #531, opened 2023-05-12
bugfix: adding EvalDelta.SharedAccts

tzaffi:
# Add missing field in `EvalDelta`

The Shared Resources work https://github.com/algorand/go-algorand/pull/5035 introduced a new field in `data.transactions.EvalDelta`. This PR adds it to the mirror type `types.EvalDelta`.

## Testing

As of 15May2023 there is a draft PR in go-algorand https://github.com/algorand/go-algorand/pull/5381 with a test that validates the types against this PR.

### DIFF
--- a/types/block.go
+++ b/types/block.go
@@ -259,6 +259,11 @@ type EvalDelta struct {
 	// [txn.Sender, txn.Accounts[0], txn.Accounts[1], ...]
 	LocalDeltas map[uint64]StateDelta `codec:"ld,allocbound=config.MaxEvalDeltaAccounts"`
 
+	// If a program modifies the local of an account that is not the Sender, or
+	// in txn.Accounts, it must be recorded here, so that the key in LocalDeltas
+	// can refer to it.
+	SharedAccts []Address `codec:"sa,allocbound=config.MaxEvalDeltaAccounts"`
+
 	Logs []string `codec:"lg"`
 
 	InnerTxns []SignedTxnWithAD `codec:"itx"`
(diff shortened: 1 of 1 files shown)


types/block.go
```diff
@@ -259,13 +259,19 @@ type EvalDelta struct {
...
+	SharedAccts []Address `codec:"sa,allocbound=config.MaxEvalDeltaAccounts"`
+
 	Logs []string `codec:"lg"`
 
 	InnerTxns []SignedTxnWithAD `codec:"itx"`
 }
 
 // StateDelta is a map from key/value store keys to ValueDeltas, indicating
 // what should happen for that key
+//
```
tzaffi, line 274:
```suggestion
```

Eric-Warehime (approved):

algochoi (commented):
Looks good - it looks like the `codec:allocbound` is defined for all the fields here in go-algorand but not here (like logs, innerTxns). Is this inconsequential?

tzaffi:
> Looks good - it looks like the `codec:allocbound` is defined for all the fields here in go-algorand but not here (like logs, innerTxns). Is this inconsequential?

Good catch. I've modified the fields in `EvalDelta` to conform with other presentations of allocbound: [EG](https://github.com/algorand/go-algorand-sdk/blob/389fc8f975301241973fb78a230001e6504d94d3/types/block.go#L277)

@Eric-Warehime - do you know if this is used programmatically in some way (you'd need reflection)? Or are these purely informational?

Eric-Warehime:
> @Eric-Warehime - do you know if this is used programmatically in some way (you'd need reflection)? Or are these purely informational?

I think it's just used by the msgp gen code i.e. https://github.com/algorand/msgp/blob/26c81f33bddb2b3897a722290e67e9e363ca96d1/parse/getast.go#L423 

So I don't think it's needed here--probably was just copied as is. But it's good to keep consistent.

algochoi (approved):

PR #534, opened 2023-05-16
lint: Add `golangci-lint` to CI process

algochoi:
Add a `lint` job to Github Actions to check for go fmt, vet and golangci-lint checks. 

The `.golangci.yml` file currently excludes generated models checks (since these cannot be hand modified easily) and checks linting errors since the v2.1 release (we will fix existing errors in another PR to make it easier for review). Otherwise, the checks were taken from [Conduit's `.golangci.yml` file](https://github.com/algorand/conduit/blob/master/.golangci.yml).

CI Job passes with no error: https://github.com/algorand/go-algorand-sdk/actions/runs/4996541721/jobs/8949856005?pr=534

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,12 +9,15 @@ linters:
   enable:
     - deadcode
     - errcheck
+    - exportloopref
     - gci
     - gofmt
     - gosimple
     - govet
     - ineffassign
     - misspell
+    - nilerr
+    - nolintlint
     - revive
     - staticcheck
     - structcheck
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,66 @@
+run:
+  timeout: 5m
+  tests: false
+  skip-dirs:
+    - client/v2/common
+  
+linters:
+  disable-all: true
+  enable:
+    - deadcode
+    - errcheck
+    - gci
+    - gofmt
+    - gosimple
+    - govet
+    - ineffassign
+    - misspell
+    - revive
+    - staticcheck
+    - structcheck
+    - typecheck
+    - unused
+    - varcheck
+
+linters-settings:
+  gci:
+    sections:
+      - standard
+      - default
+      - prefix(github.com/algorand)
+      - prefix(github.com/algorand/go-algorand-sdk)
+
+    section-separators:
+      - newLine
+
+severity:
+  default-severity: error
+
+issues:
+  # Work our way back over time to be clean against all these
+  # checkers.  If you'd like to contribute, raise the number after ~,
+  # run the linter and dig in.
+  new-from-rev: e85d1961c86009482306497313a2e0fea5bc3dbd
+
+  # Disable default exclude rules listed in `golangci-lint run --help` (selectively re-enable some below)
+  exclude-use-default: false
+
+  # Maximum issues count per one linter. Set to 0 to disable. Default is 50.
+  max-issues-per-linter: 0
+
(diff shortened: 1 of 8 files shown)


.golangci.yml
```diff
@@ -0,0 +1,64 @@
...
+    - gosimple
+    - govet
+    - ineffassign
+    - misspell
+    - revive
+    - staticcheck
+    - structcheck
+    - typecheck
+    - unused
+    - varcheck
```
tzaffi, line 23: `go-algorand` also includes the following:
```yml
linters:
...
    - exportloopref
    - nilerr
    - nolintlint
    - paralleltest
```

Chat GPT explains:
* **exportloopref**: This linter checks for pointers to enclosing loop variables. These can lead to subtle bugs due to how Go handles variable scoping and closures.
* **nilerr**: This linter checks that you are not ignoring errors returned by functions. Ignoring errors can lead to unexpected behavior.
* **nolintlint**: This linter checks for incorrect or unnecessary //nolint comments. The //nolint comment is used to tell golangci-lint to ignore certain issues on a line of code or a whole file.
* **paralleltest**: This linter checks that you are correctly using the t.Parallel() function in your tests to allow them to run in parallel.

Do you think these are worth adding?

client/v2/algod/rawTransaction.go
```diff
@@ -28,7 +28,7 @@ func (s *SendRawTransaction) Do(ctx context.Context, headers ...*common.Header)
 		}
 	}
 	if addContentType {
-		headers = append(headers, &common.Header{"Content-Type", "application/x-binary"})
+		headers = append(headers, &common.Header{Key: "Content-Type", Value: "application/x-binary"})
```
tzaffi, line 31: Is the linter requiring always using field names in struct contruction? 
There are certain situation when I explicitly want to _not_ use field names. For example, if I'm worried that another field might get added in the future and want the already defined structs to cause a compile error.

algobarb (approved):
LGTM

tzaffi (commented):
This is super!!!

Almost ready to approve. The suggested extra checks might be a pain to implement at this juncture (though I haven't tried) so I wouldn't insist on implementing them unless the current necessary fixes are manageable. 

But I am a bit opinionated on allowing structs to be constructed without field names. However, if others disagree I'll be happy to go with majority vote.

.golangci.yml
algochoi: Good suggestion, I'm in favor of adding the first three linters. I think `paralleltest` is more intrusive and caused more discussion in go-algorand. It's mostly to speed up tests by forcing parallelization, but since tests run pretty quickly in this repo, I think it would be better left out. 

Also thanks for the ChatGPT summary - saved me some clicks 😀

client/v2/algod/rawTransaction.go
algochoi: Yes, the linter currently enforces this. We can choose to exclude this rule from the check in the golanci yaml. 

I don't feel very strongly about it, but it could be nice to have backwards incompatible changes caught as a side effect. I'm not sure if the current codebase has meaningful places that would catch this though?

tzaffi: Is there a `//lint: ignore` directive that could be applied here?

algochoi: There is a directive for golangci, but none for `go vet` (it looks like go vet cannot be disabled for a particular line). We can run `go vet -composites=false ./...` to disable the rule altogether

tzaffi (approved):
LGTM

PR #538, opened 2023-05-19
algod: State delta endpoints 

algochoi:
Generate group state delta endpoints (based on this [PR](https://github.com/algorand/generator/pull/63)) and implement tests in https://github.com/algorand/algorand-sdk-testing/pull/281/files.
* /v2/deltas/{round}
* /v2/deltas/{round}/txn/group
* /v2/deltas/txn/group/{id}

We deviate slightly from the generated files - currently capitalize JSON fields such as `"Deltas"` and `"Ids"` manually

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,7 +2,7 @@ run:
   timeout: 5m
   tests: false
   skip-dirs:
-    - client/v2/common
+    - client/v2
   
 linters:
   disable-all: true
--- a/client/v2/algod/getTransactionGroupLedgerStateDeltasForRound.go
+++ b/client/v2/algod/getTransactionGroupLedgerStateDeltasForRound.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 
 	"github.com/algorand/go-algorand-sdk/v2/client/v2/common"
-	"github.com/algorand/go-algorand-sdk/v2/types"
+	"github.com/algorand/go-algorand-sdk/v2/client/v2/common/models"
 )
 
 // GetTransactionGroupLedgerStateDeltasForRoundParams contains all of the query parameters for url serialization.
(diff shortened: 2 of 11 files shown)


client/v2/algod/getTransactionGroupLedgerStateDeltasForRound.go
```diff
@@ -0,0 +1,33 @@
...
+type GetTransactionGroupLedgerStateDeltasForRound struct {
+	c *Client
+
+	round uint64
+
+	p GetTransactionGroupLedgerStateDeltasForRoundParams
+}
+
+// Do performs the HTTP request
+func (s *GetTransactionGroupLedgerStateDeltasForRound) Do(ctx context.Context, headers ...*common.Header) (response types.LedgerStateDelta, err error) {
```
winder, line 30: I'm pretty sure this is the wrong return type. It's defined as:
```
    "TransactionGroupLedgerStateDeltaForRoundResponse": {
      "description": "Response containing all ledger state deltas for transaction groups, with their associated Ids, in a single round.",
      "schema": {
        "type": "array",
        "items": {
          "$ref": "#/definitions/LedgerStateDeltaForTransactionGroup"
        }
      }
    },
```

.golangci.yml
```diff
@@ -2,7 +2,7 @@ run:
   timeout: 5m
   tests: false
   skip-dirs:
-    - client/v2/common
+    - client/v2
```
tzaffi, line 5: Is there a new warning that precipitated increasing the skip?

client/v2/algod/getTransactionGroupLedgerStateDeltasForRound.go
algochoi: Good point, looking into this

.golangci.yml
algochoi: Yes, this directory contains all the generated files (including ones from this PR in `algod.go`). It would be nice to change the generator to generate lint-compliant Go code, but I figured it was out of scope for now.

client/v2/algod/getTransactionGroupLedgerStateDeltasForRound.go
algochoi: This was fixed by adding a top level object into the API spec -- resolving

Eric-Warehime (approved):

tzaffi (approved):
LGTM

PR #543, opened 2023-05-24
types: add consensus protocol and types.

winder:
Copies the consensus types and versions from go-algorand.


tzaffi (approved):
Verified the following were identical functionally (using diff):
* `protocol/consensus.go` vs. `go-algorand/protocol/consensus.go`
* `protocol/config/consensus.go` vs. `go-algorand/config/consensus.go`

We should consider adding an x-repo-types check in `go-algorand` for `ConsensusParams` and maybe even a deeper checker that calls `ConsensusVersion()` and `initConsensusProtocols()` in each package and compares the outputs.

winder:
@tzaffi good idea. I created an issue to track that improvement:
https://github.com/algorand/go-algorand-sdk/issues/547

PR #555, opened 2023-06-08
CICD: Add github action create-release-pr workflow

algobarb:
## Summary
Add github workflow to create pull requests for releases

### DIFF
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -27,11 +27,11 @@ jobs:
         id: set-release
         run: |
           if [[ $RELEASE_VERSION =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-            if [[ $PRE_RELEASE_VERSION =~ ^[-a-z.0-9]+$ ]]; then
+            if [[ $PRE_RELEASE_VERSION =~ ^[a-z.0-9]+$ ]]; then
               echo "release-tag: $RELEASE_VERSION-$PRE_RELEASE_VERSION"
               echo "release-tag=$RELEASE_VERSION-$PRE_RELEASE_VERSION" >> $GITHUB_OUTPUT
             elif [[ -n $PRE_RELEASE_VERSION ]]; then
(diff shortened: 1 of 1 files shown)


excalq (approved):
Awesome work!

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,214 @@
...
+jobs:
+  create-release-pr:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set Release Version and Branch to Check Out
+        id: set-release
+        run: |
+          if [[ $RELEASE_VERSION =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            if [[ $PRE_RELEASE_VERSION =~ ^[-a-z.0-9]+$ ]]; then
```
onetechnical, line 30: Is this supposed to have the `-` before the a?

onetechnical: I guess that just allows dashes?

onetechnical (approved):

algobarb:
> Awesome work! (Assuming it was tested somewhere... did you do so on a fork?)

Forks from an organization repo unfortunately don't work with github actions unless you give it explicit permission to run the workflow dispatch. It said I need to be the main owner. Instead I used my own repo and copied over the files I need to update for the sdk.

algobarb:
@onetechnical Yeah it's to allow hyphens since go sdk uses hyphens for the pre-release version. So the version would be x.x.x-beta1. The dash has to be in the front or back of the regex

.github/workflows/create-release-pr.yml
algobarb: @onetechnical Yeah it's to allow hyphens since go sdk uses hyphens for the pre-release version. So the version would be x.x.x-beta.1 The dash has to be in the front or back of the regex

algobarb: Hmm... although technically I don't expect the pre_release_version to have the hyphen. I will join the `${RELEASE_VERSION}-${PRE_RELEASE_VERSION}`

algobarb: Maybe I should remove allowing a hyphen.

Issue #274, opened 2021-12-23
Add canonical check


mxmauro:
Hi, `DecodeAddress` returns the same account for the following addresses:

`PRICEP3G2F5L6ZG5WTJIAKEQW4OJJ3FM4XVFQDZI7M2VBTFVUHTTR2AU2U` (this one is the correct)
`PRICEP3G2F5L6ZG5WTJIAKEQW4OJJ3FM4XVFQDZI7M2VBTFVUHTTR2AU2V`
`PRICEP3G2F5L6ZG5WTJIAKEQW4OJJ3FM4XVFQDZI7M2VBTFVUHTTR2AU2W`
`PRICEP3G2F5L6ZG5WTJIAKEQW4OJJ3FM4XVFQDZI7M2VBTFVUHTTR2AU2X`

This issue was addressed in node's code but wondering if checking the least significant bits of the 3rd byte of the checksum to be `01` is enough as a faster approach.

Regards,
Mauro.

PR #595, opened 2023-08-21
types: Reject non canonical addresses in DecodeAddress

algochoi:
Checks that the address to decode is a canonical representation (least significant bit is 00, i.e. the ending character is one of "AEIMQUY4"). Adds a test to check that non-canonical addresses are rejected.

Note that the server (and most goal commands) will [reject non-canonical addresses](https://github.com/algorand/go-algorand/blob/master/data/basics/address.go#L75), but our SDK clients don't seem to check this.

Closes https://github.com/algorand/go-algorand-sdk/issues/274

### DIFF
--- a/types/address.go
+++ b/types/address.go
@@ -65,7 +65,8 @@ func (a *Address) UnmarshalText(text []byte) error {
 }
 
 // DecodeAddress turns a checksum address string into an Address object. It
-// checks that the checksum is correct, and returns an error if it's not.
+// checks that the checksum is correct and whether the address is canonical,
+// and returns an error if it's not.
 func DecodeAddress(addr string) (a Address, err error) {
 	// Interpret the address as base32
 	decoded, err := base32.StdEncoding.WithPadding(base32.NoPadding).DecodeString(addr)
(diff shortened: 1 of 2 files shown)


types/address.go
```diff
@@ -94,6 +95,13 @@ func DecodeAddress(addr string) (a Address, err error) {
 
 	// Checksum is good, copy address bytes into output
 	copy(a[:], addressBytes)
+
+	// Check if address is canonical
```
algochoi, line 99: https://github.com/algorand/go-algorand/blob/dfdb29324f7703f7aefa80272c7715ead75e7bb2/data/basics/address.go#L75

tzaffi (approved):
Approving as this looks very good. 

Minor nit: consider modding "correct, and ..." to "correct and canonical, and ..." in the function comment.

types/address.go
```diff
@@ -65,7 +65,8 @@ func (a *Address) UnmarshalText(text []byte) error {
 }
 
 // DecodeAddress turns a checksum address string into an Address object. It
-// checks that the checksum is correct, and returns an error if it's not.
+// checks that the checksum is correct and whether the address is canonical,
+// and returns an error if it's not.
```
tzaffi, line 69: 👍

PR #617, opened 2023-09-18
Incentives: Add fields in block header for proposer and fees collected.

jannotti:
These are needed for https://github.com/algorand/go-algorand/pull/5740 and should not be merged until that work is finalized.

### DIFF
--- a/client/v2/common/models/participation_updates.go
+++ b/client/v2/common/models/participation_updates.go
@@ -6,8 +6,4 @@ type ParticipationUpdates struct {
 	// ExpiredParticipationAccounts (partupdrmv) a list of online accounts that needs
 	// to be converted to offline since their participation key expired.
 	ExpiredParticipationAccounts []string `json:"expired-participation-accounts,omitempty"`
-
-	// AbsentParticipationAccounts contains a list of online accounts that
-	// needs to be converted to offline since they are not proposing.
-	AbsentParticipationAccounts []string `json:"absent-particpation-accounts,omitempty"`
 }
--- a/client/v2/common/models/participation_updates.go
+++ b/client/v2/common/models/participation_updates.go
@@ -6,4 +6,8 @@ type ParticipationUpdates struct {
 	// ExpiredParticipationAccounts (partupdrmv) a list of online accounts that needs
 	// to be converted to offline since their participation key expired.
 	ExpiredParticipationAccounts []string `json:"expired-participation-accounts,omitempty"`
+
+	// AbsentParticipationAccounts contains a list of online accounts that
+	// needs to be converted to offline since they are not proposing.
+	AbsentParticipationAccounts []string `json:"absent-particpation-accounts,omitempty"`
 }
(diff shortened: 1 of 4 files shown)


jasonpaulos:
I'd expect this to also update `protocol/config/consensus.go`

jannotti:
> I'd expect this to also update `protocol/config/consensus.go`

Do we just keep them in perfect sync?  Is the intent just to make everything available to callers? I can't think of any external functionality that would need access to the params, but I can see why we might have such an exposure policy. (We surely don't have it for any other sdk though)

jasonpaulos:
We only have it in this SDK for indexer & conduit to consume. There are a few consensus params that they check, such as whether rewards (the old style) are enabled, which influence their behavior/how they report certain things.

The cross repo type check in algod also verifies that the consensus params are in sync.

jasonpaulos:
> I think that's done - it's entirely manual, huh?

I thought it was a straight up copy-paste, but now I see the functions `LoadConfigurableConsensusProtocols`, `SetConfigurableConsensusProtocols`, and `PreloadConfigurableConsensusProtocols` from go-algorand are omitted. Perhaps we should move those to their own file so that it really is as simple as copy-paste (except for the "relative" import statement)

jannotti:
It was unclear to me how that model is used, but I figured I should copy the way expired accounts was done.

[a12e954](https://github.com/algorand/go-algorand-sdk/pull/617/commits/a12e95451a8b9d4239fd3268a46a50cbfa4549b8)

client/v2/common/models/participation_updates.go
```diff
@@ -6,4 +6,8 @@ type ParticipationUpdates struct {
 	// ExpiredParticipationAccounts (partupdrmv) a list of online accounts that needs
 	// to be converted to offline since their participation key expired.
 	ExpiredParticipationAccounts []string `json:"expired-participation-accounts,omitempty"`
+
+	// AbsentParticipationAccounts contains a list of online accounts that
+	// needs to be converted to offline since they are not proposing.
+	AbsentParticipationAccounts []string `json:"absent-particpation-accounts,omitempty"`
```
jasonpaulos, line 12: This file is generated, so you shouldn't manually change it

jasonpaulos: Unfortunately we don't do a good job of broadcasting that it's generated. We really should add a header saying so. Most things under the `client/v2` folder are generated by code from https://github.com/algorand/generator

This particular model is from indexer: https://github.com/algorand/indexer/blob/5ce3a09cc9e55a523a6e085930674fa9ee14fe73/api/indexer.oas2.json#L1847

The order of operations that makes the most sense would probably be merging this PR without this change, updating the indexer API, then regenerating this SDK's models based on that. Saying this because indexer will need this SDK's updated BlockHeader struct in order to populate its API response

jannotti: If I have this right, adding a field to the block header looks something like this.

1. Commit to go-algorand, despite the type check error.
2. Fix the type check error making the same changes in to go-algorand-sdk, and copy (most of) consensus.go
3. Check that, by re-running CI on go-algorand
4. Make the same changes (? not sure yet) in indexer
5. Go back and finish the changes in go-algorand-sdk, probably copying the models from indexer?

jasonpaulos (approved):

PR #618, opened 2023-12-07
Consensus Config: Add period 0 deadline timeout parameter to consensus params.

gmalouf:


protocol/config/consensus.go
```diff
@@ -1242,11 +1245,20 @@ func initConsensusProtocols() {
...
 	vFuture.EnableLogicSigCostPooling = true
 
+	vFuture.AgreementDeadlineTimeoutPeriod0 = 4 * time.Second
+
+	vFuture.StateProofBlockHashInLightHeader = true
+
+	// Setting DynamicFilterTimeout in vFuture will impact e2e test performance
+	// by reducing round time. Hence, it is commented out for now.
+	// vFuture.DynamicFilterTimeout = true
+
```
jannotti, line 1261: Is this stuff supposed to be here?  Seems like it's commenting something that isn't here.

jasonpaulos (approved):

protocol/config/consensus.go
gmalouf: It's copy pasted to match what is in go-algorand, expect to change it here in follow-up PR after we make the change in go-algorand as well.

jannotti: oh, I didn't even notice this was sdk

PR #621, opened 2023-12-15
Consensus: Update consensus files for v39 along with types sync.

gmalouf:


jasonpaulos (approved):
I believe the error can be ignored because it's to do with the outdated indexer consensus version

PR #629, opened 2024-03-07
Build: Bump go version for builds/workflows to 1.20.14

gmalouf:
Bump go version for builds/workflows to 1.20.14

Existing tests/builds should succeed.


jasonpaulos (approved):
We may need to change the PR merge rules. Looks like right now it requires the job `test_go_1.20.5` to pass, which now doesn't exist. Probably should update it to the parent job, `circleci_build_and_test `

PR #676, opened 2025-01-04
tests: cross-sdk heartbeat tests

algorandskiy:
Cross SDK tests, corresponding PR https://github.com/algorand/algorand-sdk-testing/pull/316

I had to fix the `weMakeAnyGetBlockCall` implementation to use a special mock block request that supports both JSON and MSGP. Unfortunately, there is no good way of determining the mock data content type so had to check the mocked payload for `{` symbol at the beginning to switch to JSON request.

### DIFF
--- a/test/algodclientv2_test.go
+++ b/test/algodclientv2_test.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 
 	"github.com/algorand/go-algorand-sdk/v2/client/v2/algod"
+	"github.com/algorand/go-algorand-sdk/v2/client/v2/common"
+	"github.com/algorand/go-algorand-sdk/v2/client/v2/common/models"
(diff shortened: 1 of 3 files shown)


test/algodclientv2_test.go
```diff
@@ -104,6 +107,34 @@ func weMakeAnyAccountInformationCall() error {
...
 
+type mockBlock struct {
+	c     *algod.Client
+	round uint64
+	p     algod.BlockParams
+}
+
+// This is a mock implementation of the do method in the Block struct in the algod package.
+// in order to support both JSON and MSGPACK formats depending on the mock file.
+func (s *mockBlock) do(ctx context.Context, headers ...*common.Header) (result types.Block, err error) {
```
gmalouf, line 118: Is it easier to have two separate function?

gmalouf (approved):
Asked a question, otherwise these make sense.

PR #706, opened 2025-04-16
Txns: App versioning

jannotti:
This PR was generated by a new script in go-algorand.  We should consider carefully whether the changes made are ok. The intent was to simplify things enough to make autogen easy, without any meaningful changes.

This PR intentionally does not try to adjust for the new field returned from REST APIs, as I believe that will be autogen'd once the app-version PR is merged to master.  This addresses the types that we previously edited by hand.


jannotti:
There is an unrelated box failure, but I believe this is ready for merging when the app version PR on go-algorand is merged.  It should fix the x-repo complaints. (One may remain, I don't yet understand when/how the REST API types are autogen'd, but I think that after the go-algorand merge, a PR will be generated to fix that aspect.)

gmalouf (approved):

PR #736, opened 2025-06-12
Types: Changes for tx.Access and epp consensus flag change

jannotti:
Adds fields to transaction to encode specific resources used (and consensus changes).


gmalouf (changes requested):
Discussed live: Looks like the ApplicationCallTxnFields did not get replicated over as expexted from the export-sdk-types script.

gmalouf (approved):

PR #746, opened 2025-08-29
Txn: Support lmsig logicsig signatures

jannotti:


crypto/account.go
```diff
@@ -321,12 +322,7 @@ func (lsa LogicSigAccount) IsDelegated() bool {
 func (lsa LogicSigAccount) Address() (addr types.Address, err error) {
 	hasSig := lsa.Lsig.Sig != (types.Signature{})
 	hasMsig := !lsa.Lsig.Msig.Blank()
-
-	// require at most one sig
-	if hasSig && hasMsig {
```
algorandskiy, line 326: this looked as a good early catch against malformed things

crypto/crypto.go
```diff
@@ -505,19 +508,20 @@ func VerifyLogicSig(lsig types.LogicSig, singleSigner types.Address) (result boo
...
-	// require only one or zero sig
-	if hasSig && hasMsig {
-		return false
-	}
-
-	toBeSigned := programToSign(lsig.Logic)
+	hasLMsig := !lsig.LMsig.Blank()
 
 	if hasSig {
+		if hasMsig || hasLMsig {
```
algorandskiy, line 514: maybe move it up as
```golang
if (hasSig  && hasMsig) || (hasSig && hasLMsig) || (hasMsig && hasLMsig) { return false}
```

crypto/crypto_test.go
```diff
@@ -550,15 +551,23 @@ func TestSignLogicsigTransaction(t *testing.T) {
...
+			// from: msgpacktool -e < msig_delegated.txn | base64
```
algorandskiy, line 554: 👍

crypto/account.go
gmalouf: Especially with addition of LMsig

crypto/crypto_test.go
gmalouf, line 1: Are there cases covering continued support for Msig validation?

gmalouf (dismissed):

crypto/account.go
jannotti: restored

crypto/crypto_test.go
jannotti: I added one.  `TestLogicSigMultisigLegacy`

PR #753, opened 2025-10-21
build: use Go 1.25

gmalouf:
## Summary

Bumps Go version from 1.23 to 1.25.

Fix go vet warnings preventing tests from compiling/passing due to 
https://go.dev/doc/go1.24
- "The existing printf analyzer now reports a diagnostic for calls of the form fmt.Printf(s), where s is a non-constant format string, with no other arguments. Such calls are nearly always a mistake as the value of s may contain the % symbol; use fmt.Print instead. See [#60529](https://go.dev/issue/60529)."

## Test Plan

Existing tests should pass

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,23 +20,33 @@ linters:
       require-explanation: true
       # require naming a specific linter X using //nolint:X
       require-specific: true
+    staticcheck:
+      checks:
+        - all
+        - '-ST1000' # don't require package comments
+        - '-ST1003' # don't require currentDBRound vs currentDbRound
+        - '-SA3001' # we assign to b.N in several benchmarks
+        - '-QF1008' # ignore suggestions to remove embedded fields (e.g., txn.SignedTxn.Txn -> txn.Txn)
   exclusions:
     generated: lax
     rules:
+      # Examples directory is allowed to skip error checking for brevity
+      - linters: errcheck
+        path: examples/
+      - linters: revive
+        text: '^package-comments: should have a package comment'
       # revive: types is an acceptable package name in this codebase
-      - linters:
-          - revive
+      - linters: revive
         path: types/
         text: 'var-naming: avoid meaningless package names'
-      # Examples directory is allowed to skip error checking for brevity
-      - linters:
-          - errcheck
-        path: examples/
       # Examples directory does not require comments on exported functions
-      - linters:
-          - revive
+      - linters: revive
         path: examples/
         text: 'exported: exported function .* should have comment or be unexported'
+      # ignore issues about the way we use _struct fields to define encoding settings
+      - linters: unused
+        path: (.+)\.go$
+        text: 'field _struct is unused'
       # Test utilities and helpers may have code that look unused, but is being used in another file
       - path: test/helpers.go
         text: is unused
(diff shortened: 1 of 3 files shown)


jannotti (dismissed):
Seems fine, my comments can be safely resolved.

.github/workflows/reviewdog.yml
```diff
@@ -12,17 +12,17 @@ jobs:
...
         run: go vet ./...
       - name: reviewdog-golangci-lint
-        uses: reviewdog/action-golangci-lint@v2.6.2
+        uses: reviewdog/action-golangci-lint@v2.8
         with:
-          golangci_lint_version: "v1.62.0"
+          golangci_lint_version: "v2.5.0"
           golangci_lint_flags: "-c .golangci.yml --allow-parallel-runners"
-          go_version: "1.23.3"
+          go_version: "1.25.3"
```
cce: same here, could replace with `go-version-file: 'go.mod'`

examples/apps/main.go
```diff
@@ -58,7 +58,7 @@ func appCreate(algodClient *algod.Client, creator crypto.Account) uint64 {
 		localInts   uint64 = 1
 		localBytes  uint64 = 1
 		globalInts  uint64 = 1
-		globalBytes uint64 = 0
+		globalBytes uint64
```
cce: we shouldn't let the robots boss us around

.golangci.yml
```diff
@@ -1,79 +1,102 @@
...
+      - path: (.+)\.go$
+        text: 'field _struct is unused'
+        linters:
+          - unused
+        # ignore govet false positive fixed in https://github.com/golang/go/issues/45043
+      - path: (.+)\.go$
+        text: 'sigchanyzer: misuse of unbuffered os.Signal channel as argument to signal.Notify'
+        # ignore golint false positive fixed in https://github.com/golang/lint/pull/487
+      - path: (.+)\.go$
+        text: exported method (.*).Unwrap` should have comment or be unexported
```
cce:
```suggestion
```
These should not be necessary, they were fixed a while ago and made their way into the linters

crypto/crypto.go
```diff
@@ -606,7 +606,7 @@ func SignLogicSigTransaction(lsig types.LogicSig, tx types.Transaction) (txid st
 		// the address of that account from only its signature, so assume the
 		// delegating account is the sender. If that's not the case, the signing
 		// will fail.
-		lsigAddress = tx.Header.Sender
+		lsigAddress = tx.Sender
```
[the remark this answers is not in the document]
gmalouf: Linter is updated and this change (and the 2 other embeded struct changes) were reverted.

jannotti (approved):

.golangci.yml
```diff
@@ -1,79 +1,98 @@
...
-  
 linters:
-  disable-all: true
+  default: none
   enable:
-    - errcheck
     - copyloopvar
-    - gci
-    - gofmt
-    - gosimple
```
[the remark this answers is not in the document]
gmalouf: gci and gofmt are specified further down under formatters. Typecheck and gosimple were consolidated into the staticcheck linter since last upgrade.

PR #761, opened 2025-11-24
Types: sync SDK types for on chain congestion PR

jannotti:
There will be conflicts with other changes until they are merged, but these changes sync up with the go-algorand PR to do on-chain congestion.


algorandskiy:
Replaced by https://github.com/algorand/go-algorand-sdk/pull/781

PR #794, opened 2026-04-27
TxnSubmit: Annotate API returned errors with Data field of response

jannotti:
This should expose the extra data returned when a submitted transaction fails.


jannotti:
@copilot resolve the merge conflicts in this pull request

PR #797, opened 2026-05-06
Chore: Sync for the consensus changes in big transasctions PR 6528

jannotti:
Only merge once algorand/go-algorand#6528 is merged.


algorandskiy (approved):
I've rerun against master, no changes.

PR #804, opened 2026-05-28
API: box API values, prefix, and round

joe-p:


client/v2/common/models/box_descriptor.go
```diff
@@ -1,7 +1,9 @@
 package models
 
-// BoxDescriptor box descriptor describes an app box without a value.
+// BoxDescriptor box descriptor describes an app box.
 type BoxDescriptor struct {
 	// Name base64 encoded box name
 	Name []byte `json:"name"`
+	// Value base64 encoded box value. Present only when the `values` query parameter is set to true.
+	Value []byte `json:"value,omitempty"`
```
algorandskiy, line 8: interesting why automatic tooling did not pick these up

joe-p: Yeah I'm not sure... I haven't dug too deep into how the OAS pipeline works today. Across all the SDKs its a bit of a mixed bag.

algorandskiy (approved):

PR #823, opened 2026-07-23
protocol: add fnet1..4 consensus versions

tasosbit:
Mirrors the FNet consensus versions in go-algorand ([PR here](https://github.com/algorand/go-algorand/pull/6675)) so the SDK can interpret FNet consensus params. 

Needed for conduit/indexer to work with FNet.

Verified: builds and gofmt-clean; the fnet1..4 params load into the Consensus map with the correct payout/challenge settings and upgrade edges.


jannotti:
Please hold off on considering merging this, I'm trying to generate it from go-algorand so we don't diverge.

jannotti:
See #824

PR #824, opened 2026-07-23
sync: fnet and enableF128sortition

jannotti:
This should contain #823, a recent consensus params update, and heartbeat discount field


jannotti:
@tasosbit needs this for fnet

algorandskiy (approved):

PR #825, opened 2026-07-25
feat: Support for PQ accounts

iglosiggio:
This PR adds support for PQ accounts and offers a new set of APIs for the `crypto`and `transaction` packagea.

# `crypto` package changes

- Every in-memory cryptographic operation that handled secrets was moved to a file properly marked as `_inmemorysecrets` (example: `GenerateAccount` was moved from `crypto/account.go` to `crypto/account_inmemorysecrets.go`).
- Most operations have a mirror version that takes a cryptographic signer implementation instead of using the secrets in-memory (example: `TealSign` and `Ed25519TealSign`). This allows usage of remote and/or hardware signers. Signer interfaces are `Ed25519Signer`and `PQSigner`. `PQSalted`exists only to cover surprising-non canonialized singatures.
- Added prefixed functions that take an `Ed25519Signer` instead of a raw private key. The old methods use these implementations wrapping the private key into an in-memory signer. Added `Ed25519MakeLogicSigAccountDelegated`, `Ed25519MakeLogicSigAccountDelegatedMsig`, `Ed25519AppendMultisigSignature`, `Ed25519SignTransaction`, `Ed25519SignBytes`, `Ed25519SignBid`, `Ed25519SignMultisigTransaction`, `Ed25519AppendMultisigTransaction`, `Ed25519AppendMultisigToLogicSig`, `Ed25519TealSign`, `Ed25519TealSignFromProgram`.
- Added convenience methods that transforms private keys or Ed25519 accounts into signers (`Account#AsSigner` and `SKToInMemorySigner`).
- Added signing and delegation methods for PQ signatures (`SignPQAccountTransaction` and `MakeLogicSigAccountDelegatedPQ`).
- Added convenience methods for working with PQ signers and signatures (`PQAddressFromSig`, `PQSignerAddress`, `SaltForPQSigner`).
- Added `SaltedPQSigner` a `PQSigner` implementation that allows the user to override the salt with a (maybe non-canonical) one. It implements `PQSign`, `PQPublicKey`, `PQScheme` and `PQSalt`.
- Added public constants and type definitions for `ed25519` public keys (`Ed25519PublicKeySize` and `Ed25519PublicKey`).
- Added `IsEdwards25519Point`. It is used internally for calculating the canonical salt of a PQSigner but I think it may be useful elsewhere.
- Added the extensible signer interfaces: `Ed25519Signer`, `PQSigner` and `PQSalted`. **Note on `PQSalted`**: If a `PQSigner` implementation also implements `PQSalted` then `PQSignerAddress` will return that salt, otherwise the canonical one will be calculated.
- When compiled with the `falcon` tag an in-memory account representation is included (`Falcon1024Account` that implements `Address`, `Validate` and `AsSigner`).
- When compiled with the `falcon` tag an in-memory relevant constants for falcon1024 public keys are included (`Falcon1024PublicKeySize `, `Falcon1024PublicKey`).
- When compiled with the `falcon` tag utilities for generating accounts for the in-memory representation are included (`Falcon1024AccountFromPQSeed ` and `GenerateFalcon1024Account `).
- When compiled with the `falcon` tag utilities validation of PQ signatures is enabled via `VerifyPQSig`. Note: `VerifyLogicSig` will always return `true` for PQ signatures without validating them.

# `mnemonic` package
- Updated docs description regarding in-memory secrets.
- Added `ToPQSeed` enabling a mnemonic-to-falcon-account flow.

# `transaction` package
- Added a a new collection of signer for the "extended capabilities" each signature scheme provides:
  - `DelegatableSigner` allows ed25519, multisig and pq accounts to delegate to a logicsig.
  - `Ed25519TransactionSigner` adds the capabilities currently specific to the elliptic curve scheme (`SignBytes`/`TealSign`/`AppendSignature`/`AppendDelegationSignature`).
  - Added `Ed25519AccountTransactionSigner` a `TransactionSigner` for raw ed25519 signers that is also a `DelegatableSigner` and an `Ed25519TransactionSigner`.
  - Added `MultiSigEd25519AccountTransactionSigner` a `TransactionSigner` for multisig accounts that uses raw ed25519 signers and is also a `DelegatableSigner`.
  - Added `PQAccountTransactionSigner ` a `TransactionSigner` for PQ accounts that uses raw  pq signers and is also a `DelegatableSigner`.
  - Implemented the necessary methods to make `BasicAccountTransactionSigner` a `DelegatableSigner` and an `Ed25519TransactionSigner ` (basically reimplemented `BasicAccountTransactionSigner` in terms of `Ed25519AccountTransactionSigner`).
  - Implemented the necessary methods to make `MultiSigAccountTransactionSigner ` a `DelegatableSigner` (basically reimplemented `BasicAccountTransactionSigner` in terms of `MultiSigEd25519AccountTransactionSigner`).

# `types` package
- `LogicSig#SignatureCount` now also returns `hasPQsig`. **This is a backwards-incompatible change.**
- Added `PQSchemeFalcon1024`, the protocol constant used in falcon1024 signatures and accounts.


.test-env
iglosiggio, line 1: Every change here should be removed before merging

crypto/account.go
```diff
@@ -246,6 +246,39 @@ func MakeLogicSigAccountDelegatedMsig(program []byte, args [][]byte, msigAccount
 	return
 }
 
+// MakeLogicSigAccountDelegatedPQsig creates a new delegated LogicSigAccount. This
+// type of LogicSig has the authority to sign transactions on behalf of another
+// account, called the delegating account.
+func MakeLogicSigAccountDelegatedFalcon1024(program []byte, args [][]byte, pqsigAccount Falcon1024Account) (lsa LogicSigAccount, err error) {
```
iglosiggio:
```suggestion
// MakeLogicSigAccountDelegatedFalcon1024 creates delegated LogicSigAccount that can sign on behalf of a Falcon1024 account.
func MakeLogicSigAccountDelegatedFalcon1024(program []byte, args [][]byte, pqsigAccount Falcon1024Account) (lsa LogicSigAccount, err error) {
```

crypto/crypto.go
```diff
@@ -12,6 +12,9 @@ import (
 
 	"golang.org/x/crypto/ed25519"
 
+	// NOTE: Should we move everything to the filippo dep?
+	"filippo.io/edwards25519"
```
iglosiggio: TODO: Vendor curve check

crypto/crypto.go
```diff
@@ -541,10 +544,25 @@ func VerifyLogicSig(lsig types.LogicSig, singleSigner types.Address) (result boo
...
+	}
 	// the lsig account is the hash of its program bytes, nothing left to verify
 	return true
 }
 
+// PQProgramBytesToSign returns the bytes a post-quantum scheme signs when
+// delegating a LogicSig to a PQ account: ("PQProgram" || address ||
+// program). Note that, unlike ed25519 logic-sig signing, this includes the
+// delegating account's address.
+func pqsigProgramToSign(addr types.Address, program []byte) []byte {
```
iglosiggio:
```suggestion
// pqsigProgramToSign returns the bytes a post-quantum scheme signs when
// delegating a LogicSig to a PQ account: ("PQProgram" || address ||
// program).
func pqsigProgramToSign(addr types.Address, program []byte) []byte {
```

crypto/crypto.go
```diff
@@ -625,6 +646,20 @@ func SignLogicSigTransaction(lsig types.LogicSig, tx types.Transaction) (txid st
 	return
 }
 
+func PQAddressFromSig(sig types.PQSig) (addr types.Address) {
```
iglosiggio, line 641:
```suggestion
// PQAddressFromSig returns the address of the account that performed a given PQ signature
func PQAddressFromSig(sig types.PQSig) (addr types.Address) {
```

crypto/crypto_test.go
```diff
@@ -778,3 +778,117 @@ func TestGetApplicationAddress(t *testing.T) {
...
 	require.Equal(t, expected, actual.String())
 }
+
+// The following golden tests are based on the PQ (Falcon-1024) fixtures from
+// algorandfoundation/algokit-polytest. Each reconstructs the fixture
+// transaction, re-signs it, and asserts the result equals the fixture's golden
+// signed-transaction blob byte-for-byte.
+
+// Based on pqPayment.json.
+func TestSignFalcon1024PaymentGolden(t *testing.T) {
```
iglosiggio: Use `falcon1024GoldenSeed` and `falcon1024GoldenTxn` instead.

types/transaction.go
```diff
@@ -241,21 +241,19 @@ func (tx *Transaction) Rekey(rekeyToAddress string) error {
...
+// Supported post-quantum signature schemes.
+var (
+	// PQSchemeFalcon1024 ("f1") is Falcon-1024 using a deterministic signing
+	// profile. Deterministic nonce derivation is signer-side only; consensus
+	// verifies signature validity, not how signer nonce material was derived.
+	PQSchemeFalcon1024 = PQScheme{'f', '1'}
+
+	//// PQSchemeFalcon512 ("f2") is Falcon-512 using a deterministic signing
+	//// profile. Reserved, not currently used.
+	//PQSchemeFalcon512 = PQScheme{'f', '2'}
```
iglosiggio: Should we delete this?

crypto/pqaccount.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Falcon1024Account holds both the public and private information associated with a
+// falcon address.
+type Falcon1024Account struct {
+	PublicKey  falcon.PublicKey
+	PrivateKey falcon.PrivateKey
+	Salt       types.PQAddressSalt
+}
+
+func (pqa Falcon1024Account) Address() (addr types.Address) {
```
iglosiggio:
```suggestion
// Address returns the account address for the given Falcon1024Account.
// Hash("PQA"  || scheme ||  salt || publicKey)
func (pqa Falcon1024Account) Address() (addr types.Address) {
```

crypto/pqaccount.go
```diff
@@ -0,0 +1,120 @@
...
+	buf = append(buf, pqa.PublicKey[:]...)
+
+	digest := sha512.Sum512_256(buf)
+
+	copy(addr[:], digest[:])
+
+	return
+}
+
+func (pqa Falcon1024Account) Validate() error {
```
iglosiggio:
```suggestion
// Validate returns an error if the given Falcon1024Account address could be interpreted as an ed25519 public key
func (pqa Falcon1024Account) Validate() error {
```

crypto/pqaccount.go
```diff
@@ -0,0 +1,120 @@
...
+	addr := pqa.Address()
+
+	if IsEdwards25519Point([]byte(addr[:])) {
+		return fmt.Errorf("Account address overlaps with a valid ed25519 account")
+	}
+
+	return nil
+}
+
+// Generates a new canonical falcon1024 account
```
iglosiggio:
```suggestion
// GenerateFalcon1024Account returns a new canonical Falcon1024Account
```

crypto/pqaccount.go
```diff
@@ -0,0 +1,120 @@
...
+func GenerateFalcon1024Account() (pka Falcon1024Account, err error) {
+	for {
+		seed := make([]byte, 32)
+		RandomBytes(seed)
+		pka, err = Falcon1024AccountFromPQSeed(seed)
+		if err == nil {
+			return
+		}
```
iglosiggio: Either err or no err we are already ready to return. No check needed.

crypto/pqaccount.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+func Falcon1024AccountFromPQSeed(pqseed []byte) (pka Falcon1024Account, err error) {
```
iglosiggio:
```suggestion
// Falcon1024AccountFromPQSeed returns the corresponding Falcon1024Account for a given seed.
// In conjunction with mnemonic.ToPQSeed() It can be used to generate falcon1024 accounts from regular 25 word mnemonics.
func Falcon1024AccountFromPQSeed(pqseed []byte) (pka Falcon1024Account, err error) {
```

crypto/pqaccount.go
```diff
@@ -0,0 +1,120 @@
...
+			return
+		}
+	}
+
+	pka = Falcon1024Account{}
+	err = fmt.Errorf("no valid salt with an address outside the ed25519 curve exists for the given seed")
+	return
+}
+
+func SignFalcon1024AccountTransaction(pqa Falcon1024Account, txn types.Transaction) (txid string, stxBytes []byte, err error) {
```
iglosiggio:
```suggestion
// SignFalcon1024AccountTransaction signs the given transaction with the given Falcon1024Account private key. On success it returns both transaction id and transaction bytes. 
func SignFalcon1024AccountTransaction(pqa Falcon1024Account, txn types.Transaction) (txid string, stxBytes []byte, err error) {
```

crypto/pqaccount.go
```diff
@@ -0,0 +1,120 @@
...
+	addr := pqa.Address()
+	if stx.Txn.Sender != addr {
+		stx.AuthAddr = addr
+	}
+
+	stxBytes = msgpack.Encode(stx)
+	return
+}
+
+func VerifyPQSig(toBeSigned []byte, pqsig types.PQSig) bool {
```
iglosiggio:
```suggestion
// VerifyPQSig checks that the given pqsig corresponds to the expected toBeSigned byte sequence.
func VerifyPQSig(toBeSigned []byte, pqsig types.PQSig) bool {
```

mnemonic/mnemonic.go
```diff
@@ -111,6 +117,26 @@ func ToKey(mnemonic string) ([]byte, error) {
 	return byteArr, nil
 }
 
+// ToPQSeed converts a mnemonic generated using this library into the
+// corresponding falcon PQ seed. It returns an error if the passed mnemonic
+// has an incorrect checksum, if the number of words is unexpected, or if one
+// of the passed words is not found in the words list.
+func ToPQSeed(pqmnemonic string, scheme types.PQScheme) (seed []byte, err error) {
```
iglosiggio, line 125:
```suggestion
// ToPQSeed converts a mnemonic generated using this library into the
// corresponding falcon PQ seed. It returns an error if the passed mnemonic
// has an incorrect checksum, if the number of words is unexpected, or if one
// of the passed words is not found in the words list.
// The resulting seed is Hash("PQK" || scheme || mnemonicBytes).
func ToPQSeed(pqmnemonic string, scheme types.PQScheme) (seed []byte, err error) {
```

mnemonic/mnemonic_test.go
```diff
@@ -93,3 +94,12 @@ func TestInvalidKeyLen(t *testing.T) {
 		require.Empty(t, m)
 	}
 }
+
+func TestToPQSeed(t *testing.T) {
```
iglosiggio, line 99:
```suggestion
// Based on pqMnemonic.json from algorandfoundation/algokit-polytest.
func TestToPQSeed(t *testing.T) {
```

test/steps_test.go
```diff
@@ -654,6 +662,69 @@ func equalMsigGolden(golden string) error {
 	return nil
 }
 
+// --- Falcon1024 (post-quantum) account steps. These mirror the multisig/
+// regular-account vocabulary for the operations Falcon1024 accounts support
+// (generate, import from mnemonic, address, sign, rekey). Falcon1024 is a
+// single-signer scheme, so there is no append/merge, and the signing seed is a
+// one-way hash of the mnemonic entropy so there is no mnemonic export. ---
+
```
iglosiggio:
```suggestion
```

test/steps_test.go
```diff
@@ -654,6 +662,69 @@ func equalMsigGolden(golden string) error {
...
+func genFalconKey() error {
+	var err error
+	falconAccount, err = crypto.GenerateFalcon1024Account()
+	return err
+}
+
+func loadFalconKey() error {
+	var err error
+	// 47YPQTIGQEO7T4Y4RWDYWEKV6RTR2UNBQXBABEEGM72ESWDQNCQ52OPASU
```
iglosiggio: Is this the account address? What is it supposed to represent?

transaction/transactionSigner.go
```diff
@@ -151,6 +151,43 @@ func (txSigner MultiSigAccountTransactionSigner) Equals(other TransactionSigner)
 	return false
 }
 
+type Falcon1024AccountTransactionSigner struct {
```
iglosiggio:
```suggestion
// Falcon1024AccountTransactionSigner is a TransactionSigner that can
// sign transactions for the provided Falcon1024 Account
type Falcon1024AccountTransactionSigner struct {
```

transaction/transactionSigner.go
```diff
@@ -151,6 +151,43 @@ func (txSigner MultiSigAccountTransactionSigner) Equals(other TransactionSigner)
...
+	Falcon1024Account crypto.Falcon1024Account
+}
+
+// SignTransactions signs the provided transactions with the Falcon-1024 key.
```
iglosiggio:
```suggestion
// SignTransactions signs the provided transactions with the Falcon1024 signer.
```

crypto/pqaccount.go
iglosiggio: Nope! If err is not nil we want to retry 🤦

test/steps_test.go
iglosiggio: Replaced with the actual account address.

types/signature.go
```diff
@@ -39,6 +39,38 @@ func (msig MultisigSig) Blank() bool {
...
+	}
+	if len(pqsig.PublicKey) != 0 {
+		return false
+	}
+	if len(pqsig.Signature) != 0 {
+		return false
+	}
+	return true
+}
+
```
jannotti: We should try to pull these in automatically from go-algorand, by adding to `export_sdk_types.py`.

Something like this should work:
```
export_type("PQSig", "data/transactions/pqsig.go", "signature")
export_func("(p PQSig) Blank", "data/transactions/pqsig.go", "signature")
```

I think it's good to run this after you've already got these in there (or smaller stub placeholders), and confirm that `export_sdk_types.py` replaces them with the types/func from go-algorand.

iglosiggio:
Validation for my crappy curvecheck impl is available at https://github.com/joe-p/curvecheck/pull/2

types/signature.go
iglosiggio: The other definitions on this file are not automatically exported. I'll see if the whole file can be autogenerated from go-algorand and report back :)

iglosiggio: Done at https://github.com/algorand/go-algorand/pull/6683 and pushed here :)

.test-env
iglosiggio: We have to merge https://github.com/algorand/algorand-sdk-testing/pull/326 and add [FutureDevModeNetwork.json](https://github.com/joe-p/sandbox/blob/25bad1a7f445883628deafbc3310377cf28150aa/images/algod/FutureDevModeNetwork.json#L3) to the sandbox (then make it default on the harness).

types/signature.go
iglosiggio: Alternatively I also pushed https://github.com/algorand/go-algorand/compare/export-sdk-signature-file with changes to regenerate almost all of signature.go from go-algorand (tbh some descriptions are better here, we should backport them).

types/transaction.go
iglosiggio:
```suggestion
```

crypto/crypto.go
```diff
@@ -811,3 +844,94 @@ func HashLightBlockHeader(lightBlockHeader types.LightBlockHeader) types.Digest
...
 	return sha256.Sum256(lightBlockHeaderData)
 }
+
+// IsEdwards25519Point reports whether encoded can be decoded as an
+// Edwards25519 curve point.
+//
+// Based on nullun's work: https://github.com/nullun/nullun.github.io/blob/202b399c925d51f2728c48e68d77db857a293668/curve/index.html#L291-L402
+func IsEdwards25519Point(encoded []byte) bool {
+	if len(encoded) != 32 {
+		return false
```
jannotti, line 864: Please don't reimplement cryptographic code. Use the exact same Go module go-algorand does.

iglosiggio: This was made so the go SDK didn't depend on multiple ed25519 implementation. Additionally this implementation is aligned with the impls proposed for the JS and Python SDKs.

See:
- Python impl: https://github.com/algorand/py-algorand-sdk/blob/4541936738e5f53afb1b437beb56ee1c5e41725f/algosdk/encoding.py#L215-L244
- Javascript impl: https://github.com/algorand/js-algorand-sdk/blob/b499e1e2c9b7b2fb0efae4e76c450d9aaf4da352/src/utils/ed25519-check.ts#L22-L54
- Cross-SDK validation (goal/js-sdk/python-sdk/this impl): https://github.com/joe-p/curvecheck

That said: if depending on multiple ed25519 implementations is not a problem then I'll revert the changes and use the goal reference validation.

types/signature.go
jannotti: I like the idea of exporting more, but I also ran into the concern that sometimes the SDK explanation was better, but only in the sense that it was better _for an SDK_.  If it's just plain better, let's take the SDK description.  But I recall that, for example, some LogicSIg comments were explaining how to use the SDK properly, which we shouldn't move over to go-algorand.

jannotti:
I created some merge conflicts with that last sdk update. I will fix them tomorrow if you don't get them sooner.

crypto/pqaccount.go
```diff
@@ -0,0 +1,128 @@
+package crypto
+
+import (
+	"crypto/sha512"
+	"fmt"
+
+	"github.com/algorand/falcon"
```
joe-p: In JS and Py we don't have a dependency on the falcon packages and I think it would make sense to do that here as well so we can avoid needing cgo. This would mean we don't have a `Falcon1024Account` in the SDK and instead have helper functions for going from a generic signing interface/function to a Falcon signer. 

Something similar to https://github.com/algorand/js-algorand-sdk/blob/ec2e8afa6b3941c800cde34e40cac2e0d38878bb/src/falcon-signer.ts#L18-L18

The returned signing functions can then be used in new functions that replace the ones that depend on `ed25519.PrivateKey`

types/signature.go
```diff
@@ -92,7 +117,7 @@ func (lsig LogicSig) Blank() bool {
 }
 
 // SignatureCount returns whether the LogicSig has each of the three possible signature types.
-func (lsig LogicSig) SignatureCount() (hasSig, hasMsig, hasLMsig bool, count int) {
+func (lsig LogicSig) SignatureCount() (hasSig, hasMsig, hasLMsig, hasPQsig bool, count int) {
```
iglosiggio, line 95: btw: this is a breaking change on the API.

test-harness.sh
```diff
@@ -87,9 +87,13 @@ patch_line() {
 # packages it embeds. $3 overrides the build command that install.sh runs.
 patch_sandbox_image() {
   local dir=$1 dockerfile=$2 build=${3:-make}
+  # FIXME: Remove once the sandbox is updated to reflect the sdk dep on cgo code
+  patch_line "$dir/$dockerfile" 'RUN apk add --no-cache git make bash' \
+    'RUN apk add --no-cache git make bash build-base' 'build-base'
```
iglosiggio: @joe-p the new go-algorand-sdk depends on cgo, we should pull a C toolchain on the sandbox Dockerfile then :) (sent as https://github.com/algorand/sandbox/pull/217)

joe-p (changes requested):

test-harness.sh
pbennett: Why was a cgo dependency added/needed?   This should be avoided if at all possible.

crypto/pqaccount.go
iglosiggio: I've moved everything falcon related to be gated under the "falcon" build tag. This means that no PQ support is available w/o that tag.

We can make some falcon utils dependent on a simple "pq signer" interface and make them public again (with a non-signer default impl still gated under the build tag). Do you think that's necessary or is the current state of affairs "good enough" for an initial PQ sigs release?

test-harness.sh
joe-p: It was needed for FALCON, but we are putting it behind a build tag

crypto/pqaccount.go
joe-p: Ideally the initial release of the SDKs are all aligned in terms of how they support signing. It seems preferable to delay PQ signing support a bit so we can get alignment with the other SDKs. 

I think the ideal end state is support for some like `addressWithSignersFromRawFalcon1024Signer` out of the box in the SDK and then the falcon build tag adds a layer on top that does the key gen and signing.

test-harness.sh
pbennett: Also look to https://github.com/ebitengine/purego and https://github.com/JupiterRider/ffi for ways of avoiding cgo.  https://github.com/hybridgroup/yzma/tree/main uses them to avoid cgo usage.

crypto/crypto.go
```diff
@@ -541,10 +542,24 @@ func VerifyLogicSig(lsig types.LogicSig, singleSigner types.Address) (result boo
 		toBeSigned := msigProgramToSign(addr, lsig.Logic)
 		return VerifyMultisig(addr, toBeSigned, lsig.LMsig)
 	}
+
+	if hasPQsig && lsig.PQsig.Scheme == types.PQSchemeFalcon1024 {
```
joe-p: We discussed what to do with the verify functions with the protocol team and the decision was to just deprecated them. For this reason I think we should just remove this conditional entirely and note that this is not a reliable method generally and specifically for PQ

crypto/pqaccount_nofalcon.go
```diff
@@ -0,0 +1,12 @@
+//go:build !falcon
+
+package crypto
+
+import "github.com/algorand/go-algorand-sdk/v2/types"
+
+// verifyPQSig is the unexported version of VerifyPQSig used by VerifyLogicSig.
+// On falcon-enabled environment the implementation will use the falcon impl to validate the signature.
+// See pqaccount_falcon.go for more info.
+func verifyPQSig(toBeSigned []byte, pqsig types.PQSig) bool {
```
joe-p: Since we are deprecating verification functions we can probably just remove this function all together. It also seems nice to avoid one function behaving differently depending on build tag

types/signature.go
joe-p: We added lmsig and no one seemed to complain, so I think this is fine

joe-p (changes requested):
LGTM. I think we should just remove verify entirely. Also we should think about what the plan for 512 is. Will we just replicate everything we have for 1024?

joe-p:
Thinking about this a bit more... I think it will make more sense to have the crypto agnostic signing interfaces first and then we can have a falcon generator behind the falcon build tag. This avoids us effectively needing to implement all the PQ signing functions twice (once for `Falcon1024Account` and another for the agnostic signing interfaces)

README.md
```diff
@@ -8,6 +8,9 @@ The Algorand golang SDK provides:
 - HTTP clients for the algod (agreement) and kmd (key management) APIs
 - Standalone functionality for interacting with the Algorand protocol, including transaction signing, message encoding, etc.
 
+> [!NOTE]
+> Support for falcon signatures is gated under the `falcon` tag since it requires using the [algorand/falcon](https://github.com/algorand/falcon) cgo library.
```
iglosiggio: I should rephrase this.

test-harness.sh
iglosiggio: For now we decided to put the falcon dependency under a build tag.

Regular builds will get an API that requires a PQ signer implementation but no in-memory implementation will be provided unless you use the `falcon` tag.

joe-p:
I don't love having functions like `Ed25519SignTransaction` exported, but I understand we need to export them because they are used by the transaction package. What if we just moved everything to the transaction package so we can keep the scheme-specific functions private? I would say this actually better aligns with the package naming because there isn't any crypto actually being done by these functions (they are simply giving the signers the correct shape), so keeping the `crypto` package purely for cryptographic operations would make sense.

joe-p (approved):
Current state seems good to me. @iglosiggio I have added three commits since we last spoke. The commit messages contain rationale for the changes. 

[8958274](https://github.com/algorand/go-algorand-sdk/pull/825/commits/89582746844180235c712d058c0f7e6eb74b8bda)

[d47e155](https://github.com/algorand/go-algorand-sdk/pull/825/commits/d47e15508c92eae852ec596bdd9b67d86b7c9c3b)

[8958274](https://github.com/algorand/go-algorand-sdk/pull/825/commits/89582746844180235c712d058c0f7e6eb74b8bda)

PR #826, opened 2026-07-29
protocol: add fnet5 consensus version

tasosbit:
fnet5 is the protocol AF's FNet runs. It extends v41 with the parameter changes go-algorand released as ConsensusV42, restated rather than derived from v42 (in case it changes before it is finalized), so fnet5 stays pinned to the values FNet runs with.

Also adds the V41 -> fnet5 upgrade path.


algorandskiy:
reopening to retrigger CLA

algorandskiy (approved):

PR #828, opened 2026-07-29
Sync: Update for v42 consensus

jannotti:
I think this will be the wrong place for PQSig eventually, but the PR moving it to `signature.go` hasn't landed yet, so this hopefully minimizes changes`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # ignore cucumber test resources
 test-harness/
+.staged-sandbox/
 test/features/
 test/features/resources/**/*.json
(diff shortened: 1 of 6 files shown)


algorandskiy (dismissed):

iglosiggio:
PR has landed, should we re-sync?

jannotti:
> PR has landed, should we re-sync?

I'll adjust this one.

jannotti:
Wait, I'm confused.  I was referring to #825 which still is not in.

algorandskiy (approved):

PR #831, opened 2026-08-01
fix: make API error wrappers distinguishable at runtime

### LINKED ISSUE #501: Fix API Error Wrappers
### Subject of the issue

The way API errors are wrapped does not work. Having type aliases to an interface means that at runtime all of the errors are indistinguishable from one another.

Here is the current code:
```
type BadRequest error
type InvalidToken error
type NotFound error
type InternalError error

// extractError checks if the response signifies an error.
// If so, it returns the error.
// Otherwise, it returns nil.
func extractError(code int, errorBuf []byte) error {
	if code >= 200 && code < 300 {
		return nil
	}

	wrappedError := fmt.Errorf("HTTP %v: %s", code, errorBuf)
	switch code {
	case 400:
		return BadRequest(wrappedError)
	case 401:
		return InvalidToken(wrappedError)
	case 404:
		return NotFound(wrappedError)
	case 500:
		return InternalError(wrappedError)
	default:
		return wrappedError
	}
}
```

### Solution

Create a real error and use that for the specific types. They should also include the error code, for example:
```
// You can edit this code!
// Click here and start typing.
package main

import "fmt"

type HTTPError struct {
	Msg  string
	Code uint
}

func (e HTTPError) Error() string {
	return fmt.Sprintf("Client error (%d): %s", e.Code, e.Msg)
}

type NotFound HTTPError

func (e NotFound) Error() string {
	return fmt.Sprintf("NotFound error (%d): %s", e.Code, e.Msg)
}

func main() {
	genericError := HTTPError{
		Msg:  "Method Not Found",
		Code: 404,
	}

	fmt.Printf("Error type: %T, Error message: %v\n", genericError, genericError)

	var notFound NotFound = NotFound{
		Msg:  "Method Not Found",
		Code: 404,
	}
	fmt.Printf("Error type: %T, Error message: %v\n", notFound, notFound)
}
```

Create types so that the `extractError` can be updated as follows:
```
// extractError checks if the response signifies an error.
// If so, it returns the error.
// Otherwise, it returns nil.
func extractError(code int, errorBuf []byte) error {
	if code >= 200 && code < 300 {
		return nil
	}

	switch code {
	case 401:
		return InvalidToken{Msg: errorBuf}
	case 404:
		return NotFound{Msg: errorBuf}
	case 500:
		return InternalError(wrappedError)
	default:
		if code < 400 {
			return ClientError{Code: code, Msg: errorBuf}
		} else if code < 500 {
			return ServerError{Code: code, Msg: errorBuf}
		}
		return UnknownError{Code: code, Msg: errorBuf}
	}
}
```

mitre88:
## Summary
- Replaces `type BadRequest error` (and siblings) with concrete structs that implement `error`.
- Each typed error includes `StatusCode` and `Message`, so `errors.As` / type assertions work at runtime.
- Adds `HTTPError` for non-mapped status codes.
- Extends unit tests to assert runtime type distinguishability.

## Context
Fixes #501. Type aliases to the `error` interface are indistinguishable at runtime; callers cannot reliably switch on API failure class.

## Validation
```bash
go test ./client/v2/common/ -count=1
```

### DIFF
--- a/client/v2/common/common.go
+++ b/client/v2/common/common.go
@@ -78,10 +78,55 @@ func MakeClientWithTransport(address string, apiHeader, apiToken string, headers
 	return
 }
 
-type BadRequest error
-type InvalidToken error
-type NotFound error
-type InternalError error
+// HTTPError is returned for non-2xx HTTP responses that do not map to a more specific type.
+type HTTPError struct {
+	StatusCode int
+	Message    string
+}
+
+func (e HTTPError) Error() string {
+	return fmt.Sprintf("HTTP %v: %s", e.StatusCode, e.Message)
+}
+
+// BadRequest is returned for HTTP 400 responses.
+type BadRequest struct {
+	StatusCode int
+	Message    string
+}
+
+func (e BadRequest) Error() string {
+	return fmt.Sprintf("HTTP %v: %s", e.StatusCode, e.Message)
+}
+
+// InvalidToken is returned for HTTP 401 responses.
+type InvalidToken struct {
+	StatusCode int
+	Message    string
+}
+
+func (e InvalidToken) Error() string {
+	return fmt.Sprintf("HTTP %v: %s", e.StatusCode, e.Message)
+}
+
+// NotFound is returned for HTTP 404 responses.
+type NotFound struct {
+	StatusCode int
+	Message    string
+}
+
+func (e NotFound) Error() string {
+	return fmt.Sprintf("HTTP %v: %s", e.StatusCode, e.Message)
+}
+
+// InternalError is returned for HTTP 500 responses.
+type InternalError struct {
+	StatusCode int
+	Message    string
+}
+
+func (e InternalError) Error() string {
+	return fmt.Sprintf("HTTP %v: %s", e.StatusCode, e.Message)
+}
 
 // extractError checks if the response signifies an error.
 // If so, it returns the error.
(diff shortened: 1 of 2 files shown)


mitre88:
Hi @jannotti @algorandskiy, this PR is ready for review. It fixes #501 by replacing `type BadRequest error` aliases with concrete structs, so `errors.As` and type assertions work at runtime. Unit tests cover runtime type distinguishability.

One heads up: the CLA assistant badge still shows not signed, so it may need a recheck after the contributor signs. The commit is ready to review in the meantime.

client/v2/common/common.go
```diff
@@ -78,10 +78,55 @@ func MakeClientWithTransport(address string, apiHeader, apiToken string, headers
...
+// BadRequest is returned for HTTP 400 responses.
+type BadRequest struct {
+	StatusCode int
+	Message    string
+}
+
+func (e BadRequest) Error() string {
+	return fmt.Sprintf("HTTP %v: %s", e.StatusCode, e.Message)
+}
+
```
jannotti, line 95: Having introduced HTTPError above, can the repetition below be reduced by embedding HTTPError in each newly introduced error? You get the field definitions and the Error() definition for free.

mitre88: Done — BadRequest, InvalidToken, NotFound and InternalError now embed HTTPError, dropping the duplicated fields and Error() methods. Type assertions still work since each keeps its own concrete type. Diff is now +12/-32 across the two files.

jannotti (approved):

jannotti:
I like it now. @joe-p , what do you think?

joe-p:
Yeah this seems better but it is a breaking change. If we are going to make a breaking change to the API clients it might be worthwhile to see if there other changes we want to make as well.

jannotti:
I don't think it's breaking in any meaningful way.  There was no useful way
to use `Error.As()` before (because each error was an `error`), so I don't
think this is user visible.

algorandskiy (approved):

mitre88:
Hi @jannotti — thanks for the review and approval! 🙏

Quick status update on this PR:

- The two commits were re-signed with SSH keys (GitHub now shows them as **Verified**): `231a53a` and `c59ab39`.
- The three workflow runs on the new head (`c59ab39`) are stuck in `action_required` because this is a fork PR — they need a maintainer to click **Approve and run** once.
- `license/cla` already passed (`success`).

Once the runs are approved and pass, the PR should be mergeable. Could you approve the 3 runs (Test Suite, Lint Checks, Check PR category and type) and merge when green? Happy to address any feedback.

Thanks again for the review!

mitre88:
Hi maintainers — gentle nudge on this one. It has two approvals and the CLA check passed, and it's currently mergeable with no failing checks.

Just checking: is anything else needed before this can be merged? Happy to rebase or adjust if there's something outstanding. Thanks!

joe-p:
Sorry for the delay. We've been rather busy with the upcoming consensus update. Thanks for the contribution!

P.S I have seen your other PRs in various repos. I hope to get to them all by the end of this week.

PR #836, opened 2026-08-12
feat: Add an extensible signer API

iglosiggio:
This PR adds the interface `crypto.Ed25519Signer` and replaces every usage of `ed25519.PrivateKey` with an implementation that depends on the signer.

The previous API surface is kept on `_inmemory.go` files meanwhile the new API uses `Ed25519` as a prefix to differentiate them. The previous API is implemented in terms of the new one and all tests are migrated to the new API.


.test-env
iglosiggio, line 1: This should be reverted before merging (is only here because we are currently failing the falcon tests)

iglosiggio:
Added falcon signers and modified the API to be only in the "new style"

joe-p (changes requested):
Rather than making the functions Ed25519-specific they should be agnostic to the underlying crypto scheme and instead use domain-specific signers. 

For example, 

```go
Ed25519MakeLogicSigAccountDelegated(program []byte, args [][]byte, signer Ed25519Signer) (lsa LogicSigAccount, err error)
```

Would be

```go
MakeLogicSigAccountDelegated(program []byte, args [][]byte, signer DelegatedLogicSigSigner) (lsa LogicSigAccount, err error)
```

Then there can be an entry-point that takes a `Ed25519Signer` and returns the `DelegatedLogicSigSigner`. For example, in TypeScript:

```ts
export interface Ed25519SigningKey {
  ed25519PublicKey: Uint8Array;
  ed25519Signer: (bytesToSign: Uint8Array) => Promise<Uint8Array>;
}

export function addressWithSignersFromRawEd25519Signer(
  ed25519SigningKey: Ed25519SigningKey,
  sendingAddress: Address = new Address(ed25519SigningKey.ed25519PublicKey)
): AddressWithTransactionSigner &
  AddressWithDelegatedLsigSigner &
  AddressWithMxBytesSigner &
  AddressWithProgramDataSigner 
```

As we add additional schemes and signing operations, we simply need to update this single point of entry.

iglosiggio:
> Rather than making the functions Ed25519-specific they should be agnostic to the underlying crypto scheme and instead use domain-specific signers.
> 
> For example,
> 
> ```go
> Ed25519MakeLogicSigAccountDelegated(program []byte, args [][]byte, signer Ed25519Signer) (lsa LogicSigAccount, err error)
> ```
> 
> Would be
> 
> ```go
> MakeLogicSigAccountDelegated(program []byte, args [][]byte, signer DelegatedLogicSigSigner) (lsa LogicSigAccount, err error)
> ```
> 
> Then there can be an entry-point that takes a `Ed25519Signer` and returns the `DelegatedLogicSigSigner`. For example, in TypeScript:
> 
> ```ts
> export interface Ed25519SigningKey {
>   ed25519PublicKey: Uint8Array;
>   ed25519Signer: (bytesToSign: Uint8Array) => Promise<Uint8Array>;
> }
> 
> export function addressWithSignersFromRawEd25519Signer(
>   ed25519SigningKey: Ed25519SigningKey,
>   sendingAddress: Address = new Address(ed25519SigningKey.ed25519PublicKey)
> ): AddressWithTransactionSigner &
>   AddressWithDelegatedLsigSigner &
>   AddressWithMxBytesSigner &
>   AddressWithProgramDataSigner 
> ```
> 
> As we add additional schemes and signing operations, we simply need to update this single point of entry.

See https://github.com/algorand/go-algorand-sdk/pull/836/changes/dac5188c7f5d08ecdc60aeb0850666ab43703fad for my take on this.

I've also addressed some of the comments for #825 here. Should we merge everything into a single PR?

PR #838, opened 2026-08-13
fix: reject asset 0 in Access foreign assets

### LINKED ISSUE #764: Using AtomicTransactionComposer.AddMethodCall w/ UseAccess set and []ForeignAssets w/ 0 isn't rejected
### Your environment

* go-algorand-sdk: 2.11.1

### Steps to reproduce

1. Construct a transaction.AddMethodCallParams{ ..} instance w/ AppID, method args, etc.
2. Set UseAccess: true
3. Add 0 as member of ForeignAssets slice.
4. Call AddMethodCall(params) on AtomicTransactionComposer instance.

### Expected behaviour

A foreign asset reference to 0 isn't valid and should return an error from AddMethodCall

### Actual behaviour

The access list composition code adds an empty reference for the asset 0 reference, creating what is technically 'valid', but different, an empty ResourceRef is really treated at runtime by the AVM as an empty box reference.  Callers should have to pass an empty box reference to do this, not a 0 asset reference.

mitre88:
Fixes #764.

`AtomicTransactionComposer.AddMethodCall` with `UseAccess: true` previously accepted `ForeignAssets` containing `0`. An all-zero `ResourceRef` is encoded as an empty box, not an asset, so the AVM treated that reference as an empty box.

This rejects asset ID `0` when composing an Access list (foreign assets and holdings that would otherwise become an empty `ResourceRef`). Callers who want an empty box must pass an empty box reference.

Legacy foreign-array application calls without `UseAccess` are unchanged; asset `0` remains allowed there.

## Test
`go test ./transaction/` — pass


mitre88:
Rebased onto current `main` (v2.12.0 / consensus v42). Local `go test ./transaction/` pass. CLA is green; GitHub Actions on this fork still need a first-time workflow approval to run.

jannotti (approved):

PR #845, opened 2026-09-10
refactoring pqsig-support

joe-p:
Primarily moves signing stuff from crypto to transaction since there isn't any crypto being done and lets us share stuff within transaction. Some Ed-specific functions need to be exported because of package boundaries but that is preferable to lots of duplication. The tests and examples, however, were updated to use the scheme agnostic interfaces

Also includes various small fixes that were found during LLM code review


crypto/errors.go
```diff
@@ -4,7 +4,14 @@ import (
 	"errors"
 )
 
-var errInvalidSignatureReturned = errors.New("ed25519 library returned an invalid signature")
+// ErrInvalidSignatureReturned is returned when an Ed25519Signer produces a
+// signature whose length is not ed25519.SignatureSize.
+var ErrInvalidSignatureReturned = errors.New("ed25519 library returned an invalid signature")
```
iglosiggio: I don't think narrowing the meaning of "invalid" helps here (also: why is it a public symbol?).

crypto/errors.go
```diff
@@ -4,7 +4,14 @@ import (
...
+
+// ErrNilPQSigner is returned when a post-quantum signing operation is given no
+// signer to sign with.
+var ErrNilPQSigner = errors.New("pq signer cannot be nil")
```
iglosiggio: Why is it public?

crypto/pqaccount.go
```diff
@@ -61,110 +31,97 @@ func pqAddress(pk []byte, scheme types.PQScheme, salt types.PQAddressSalt) (addr
...
+func PQSignerAddress(signer PQSigner) (addr types.Address, err error) {
+	if signer == nil {
+		return types.Address{}, ErrNilPQSigner
+	}
+	return PQAddress(signer.PQPublicKey(), signer.PQScheme())
+}
+
+// SaltForPQSigner returns the canonical salt that will be used when performing
+// PQ signatures on behalf of the given signer.
 func SaltForPQSigner(sgnr PQSigner) (types.PQAddressSalt, error) {
```
iglosiggio: Please don't. If you want to have something that returns the canonical salt make it `CanonicalSaltForPQSigner`.

Saying that this will always returns the canonical salt puts us in an ugly position if at some point in the future we want to support non-canonical salts.

crypto/pqaccount.go
```diff
@@ -61,110 +31,97 @@ func pqAddress(pk []byte, scheme types.PQScheme, salt types.PQAddressSalt) (addr
...
-	if salted, ok := sgnr.(PQSalted); ok {
-		return salted.PQSalt(), nil
+	if sgnr == nil {
+		return 0, ErrNilPQSigner
 	}
```
iglosiggio: Why are we checking this? I would have assumed that "gimme a signer and I will give you its salt" already requires that you give it a signer.

crypto/pqaccount.go
```diff
@@ -61,110 +31,97 @@ func pqAddress(pk []byte, scheme types.PQScheme, salt types.PQAddressSalt) (addr
...
-	}
-
-	pk := sgnr.PQPublicKey()
+// PQSigFor assembles the PQSig envelope that identifies the signer's account
+// and carries the given signature bytes, alongside the address of that account.
+//
+// The signature is not checked, and may be empty: callers that need an envelope
+// without a real signature (to size or simulate a transaction, say) can pass
+// nil.
+func PQSigFor(sgnr PQSigner, signature []byte) (sig types.PQSig, addr types.Address, err error) {
```
iglosiggio: This function feels pretty weird. It just shuffles data around and forces the caller code to make no mistakes (or irrelevant mistakes, for the simulation case).

crypto/pqaccount.go
```diff
@@ -61,110 +31,97 @@ func pqAddress(pk []byte, scheme types.PQScheme, salt types.PQAddressSalt) (addr
...
+	}
 	if err = sanityCheckProgram(program); err != nil {
 		return
 	}
 
-	pk := sgnr.PQPublicKey()
-	salt, err := SaltForPQSigner(sgnr)
+	// the delegation signature commits to the address it delegates from, so the
+	// envelope is assembled first and its signature filled in afterwards
+	pqsig, addr, err := PQSigFor(sgnr, nil)
```
iglosiggio: This feels really weird. Doesn't it make writing the wrong code a lot easier?

types/signature.go
```diff
@@ -88,11 +88,14 @@ func (lsig LogicSig) Blank() bool {
...
+	if !lsig.PQsig.Blank() {
+		return false
+	}
 	return true
 }
 
-// SignatureCount returns whether the LogicSig has each of the three possible signature types.
-func (lsig LogicSig) SignatureCount() (hasSig, hasMsig, hasLMsig, hasPQsig bool, count int) {
+// SignatureCount returns whether the LogicSig has each of the three possible classical signature types.
+func (lsig LogicSig) SignatureCount() (hasSig, hasMsig, hasLMsig bool, count int) {
```
iglosiggio: This is also weird.

